PR #7, opened 2021-12-03
fix: link to jet package

dylancom:
### Description

Error upon installing the project: Couldn't find match for "ff3f9bec4a53f7f1e1c05dcc97abede26c72e72f"


mikehardy (approved):
Yeah this one is going to be unstable unfortunately, need to get jet-next out - tangled with this one - before it will stabilize. But this should work for now. Yarn has a behavior where if you depend on a tag or branch it will not update the commit hash it *thinks* is the concrete reference to the named git item unless you re-add it locally, but there's no way around it and this will be more stable.

PR #8, opened 2021-12-03
fix: missing dollar sings

dylancom:
### Description

Setting $RNGoogleAdsSDKVersion and/or $RNGoogleAdsAsStaticFramework is currently not working.


mikehardy (approved):
Ah, that's a :facepalm:  one - I wasn't happy publishing it in the state it was because of things like this, but better out and ugly so it has a hope of getting fixed then closed :shrug:  - thanks for fixing this

PR #14, opened 2021-12-05
fix: displaying banner ads, wrong config in docs

dylancom:
### Description

I just initialized a new RN project, followed the docs and noticed the app would crash at first.
The crash was because the google ads sdk couldn't find the app_id from the app.json config.
Therefore I updated the docs.

Then I noticed banners would still not render, one more name change was required (which I think I forgot to push in my previous PR's).


mikehardy (approved):
Nice - yes the naming part of the extraction touched everything and is probably still wrong in a few corners, e.g. the android side may have an inconsistent module name still but this is all clearly improvement, thanks again!

mikehardy:
Ha just realized you only posted this 4 minutes ago. I need to be async again today but you just happened to post this while I was doing my morning coffee + github notification scan ;-)

dylancom:
Haha, amazing timing!

PR #17, opened 2021-12-05
fix: banner ads not working on android

dylancom:
### Description

Banners were not working because of a naming mismatch.


mikehardy (approved):
Yep, as expected :sweat_smile: - hopefully it's the last of the naming mismatches - have you grepped through the whole codebase or are you doing them one at a time?

dylancom:
Had only time to test banners for now. For sure there are more changes needed for the other formats.

mikehardy:
Let's just get them instead of going piecemeal. It'll be faster.

```
mike@bistromath:~/work/Invertase/react-native-google-ads (main) % grep -r ReactNativeGoogleAds *|grep REACT_CLASS
android/src/main/java/io/invertase/googleads/ReactNativeGoogleAdsBannerAdViewManager.java:  private static final String REACT_CLASS = "ReactNativeGoogleAdsBannerView";
mike@bistromath:~/work/Invertase/react-native-google-ads (main) % 
```

looks like it's just that one on android side at least

PR #19, opened 2021-12-07
[proposal] Renaming rnapp

dylancom:
### Description

This is a small proposal to rename RNApp to RNGAApp (which used to be RNFBApp).


mikehardy:
wrong branch I think :-), I see you closed it immediately - but philosophically, I want everything in "common" to be pure of module-specific stuff. It will be picked up and used immediately for MLKit and moved back to RNFB after it gets battle-tested here and MLKit as a more generic piece of infrastructure, ideally as a library or something so it's by reference

PR #23, opened 2021-12-08
fix: jest task fails

dylancom:
### Description

This fixes the jest task

### DIFF
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -9,6 +9,10 @@ jest.doMock('react-native', () => {
       },
       NativeModules: {
         ...ReactNative.NativeModules,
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
thanks - ran out of time on that last night, it was nearly 3am here :weary: 
getting really close I just wanted to have a certain baseline of CI assurance that things weren't going completely off the rails so any PRs can be merged with some confidence

mikehardy:
What is really puzzling to me - where I got stuck before I ran out of time - is that react-native is mocked so that Platform.OS should return android, meaning the assertion that was throwing should not be possible as that requires a Platform of iOS, but who knows :shrug:

PR #26, opened 2021-12-09
fix: incorrect rn config path

dylancom:
### Description

Currently the google-ads config in app.json is ignored because the path to react-native.config.js is incorrect.


mikehardy (approved):
All mine, that one.thanks

PR #30, opened 2021-12-10
feat(ios)!: Google Mobile Ads SDK 8 w/new testDeviceIdentifiers vs testDevices

dylancom:
### Changes
- testDevices became testDeviceIdentifiers and have to be set in the Admob request configuration instead of per ad.
- Smart banner deprecation in favor of adaptive banner.
- Leave application callback removal
The willLeaveApplication callback for all ad formats has been removed in favor of the applicationDidEnterBackground: and sceneDidEnterBackground: methods. Using OS-level APIs notify you whenever users leave your app, regardless of whether or not it is due to an ad interaction.
Note that the willLeaveApplication callback was never intended to be an ad click handler, and relying on this callback to report clicks did not produce an accurate metric. For example, a click on the AdChoices icon that launched an external browser invoked the callback but did not count a click.


lib/index.d.ts
```diff
@@ -669,19 +669,6 @@ export namespace GoogleAdsTypes {
...
-     *
-     * If using an emulator, set the device ID to `EMULATOR`.
-     *
-     * ```js
-     * await Interstitial.request('ca-app-pub-3940256099942544/1033173712', {
-     *   testDevices: ['EMULATOR'],
-     * });
-     * ```
-     */
-    testDevices?: string[];
```
mikehardy, line 683: @dylancom this is the only part I don't understand - it seems like it's performing the exact  same job and has the same typing, but if you change the name from testDevices to testDeviceIdentifiers it becomes a breaking change. Is that on purpose? Is it desired? :thinking:

dylancom: Yes so, now you set this option once instead of per each ad type (if you use banners/interstials etc in 1 app). At first I thought to leave the name intact, but then I thought about aligning it with the naming used in the sdk. (As they did the same name change there). Can easily keep the original name though if desired.

dylancom: "The testDeviceIdentifiers property applies to all ad requests, while the old testDevices property was set per-request."

mikehardy: Sounds like a breaking change in behavior, probably best to go ahead and break the code. Let's test that semver and get a v2 ;-)

Serves as a good time to let you know my philosophy in general - copied from webview: 

> This project follows semantic versioning. We do not hesitate to release breaking changes but they will be in a major version.

Breaking change happens, no problem. Just bump that major. Vamos!

mikehardy:
@dylancom :warning: :warning: I rebased you from main where I *think* I fixed the patch-package script, and I ran `yarn lint:ios:fix` - with those I think this will actually be good to merge and CI will be all green + patches :crossed_fingers: 
I know it's Friday night late there, I *hope* this wasn't too bold, I couldn't imagine you were still hacking on it? I'm about to be off for the night myself. Cheers

mikehardy:
Patches work!!! https://github.com/invertase/react-native-google-ads/actions/runs/1565648878

mikehardy:
I hopefully fixed release automation's final step with a new token. PR title changed to signify breaking status (`!` w/brief explanation). All queued up on the runway ready to lift off

mikehardy:
Yep, new github PAT did the trick. We are 100% automated now. Just missing on-device tests and the slow battle to get coverage back up. It'll get there. Thanks so much for getting this package going, it means a lot - it was really weighing on me as RNFB maintainer, having made some promises to get it going and then not having the time. Big weight off my shoulders. Cheers

PR #33, opened 2021-12-11
docs: improve outbound requests info

dylancom:
### Description

Currently the docs make it seem like setting setRequestConfiguration is mandatory.
Actually it's only needed if your app requires specific targeting or you need to provide testID's.

### DIFF
--- a/docs/index.mdx
+++ b/docs/index.mdx
@@ -70,13 +70,12 @@ npx react-native run-android
 
 ## Configure outbound requests
 
-Before requesting ads, you must tell the AdMob network what type of content you'd wish to receive based on your target
-audience.
+If needed, you can provide settings that will apply to all ad requests.
 
 For example, if the application targets children then you must configure the outbound requests to only
 receive content suitable for children before loading any adverts.
 
-To set the request configuration, call the `setRequestConfiguration` method:
+To set the request configuration, call the `setRequestConfiguration` method before requesting ads:
 
 ```js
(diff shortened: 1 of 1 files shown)


docs/index.mdx
```diff
@@ -70,13 +70,12 @@ npx react-native run-android
 
 ## Configure outbound requests
 
-Before requesting ads, you must tell the AdMob network what type of content you'd wish to receive based on your target
-audience.
+If needed, you can provide settings that will apply to all ad requests.
```
mikehardy, line 73:
```suggestion
If the default ad settings are not correct for your app, you can provide settings that will apply to all ad requests.
```

docs/index.mdx
```diff
@@ -70,13 +70,12 @@ npx react-native run-android
...
 
 For example, if the application targets children then you must configure the outbound requests to only
 receive content suitable for children before loading any adverts.
 
-To set the request configuration, call the `setRequestConfiguration` method:
+To set the request configuration, call the `setRequestConfiguration` method before requesting ads:
```
mikehardy, line 78:
```suggestion
If you need to set a custom request configuration, you must call the `setRequestConfiguration` method before requesting ads:
```

mikehardy (commented):
Nice!

just some wording changes to clarify even more that defaults exist, and this is just to go custom
and then to emphasize that if you go custom, it must be before requests

what do you think?

mikehardy (approved):

Issue #4, opened 2021-09-20
Initialize/Start not called


TaylorDale:
Hi again - apologies for polluting the new repo with issues... 

It would be very helpful if there was a method to call initialize() (android) and start() (iOS) on the Admob SDK. Admob ads work fine if this is not called, but most mediation (AdColony, FAN for example) will not work fullstop if these respective functions are not called (Worth noting, Unity ads *does* work without calling this but that is the only one I've found so far). The mediation test suite also will fail to launch if these haven't been called. It makes for a really hard-to-debug mediation situation as when init() isn't called there is not much to tell you that is the problem.

I've just ended up placing these methods in my AppDelegate onLaunch/Activity OnStart methods but it would be great to be able to call them from RN.

mikehardy:
Very interesting! Will keep this in the queue. I've been traveling which put the extraction/bring-up of this code from react-native-firebase on hold for a week or so but I'm about to get back to it this week. First step will be a straight extraction / re-vivification with no breaking changes or features other than symbol names moving around.

Then I'll forward port to new SDKs

Then I'll scan the list as things come in here (like this one) and try to get to them, or perhaps you can help with a PR once the repo is actually alive and kicking 🙏 😄

TaylorDale:
Awesome thanks for the hard work put in to these libs. Would love to submit a PR for this one - will see if I get some free time.

mikehardy:
@TaylorDale these are about to launch...

TaylorDale:
@mikehardy thanks mate appreciate all the hard work!

PR #40, opened 2021-12-13
feat: expose native SDK initialize methods as new `initialize` API

dylancom:
### Description

It has been adviced to initialize the sdk before loading ads. (Especially when using mediation).
Therefore these methods should be exposed to RN.

Solves: #4

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -726,15 +726,13 @@ export namespace GoogleAdsTypes {
   }
 
   /**
-   * Initialization status of each ad network available to the Google Mobile Ads SDK, keyed by its
-   * GADMAdapter's class name. The list of available ad networks may be incomplete during early
-   * phases of SDK initialization.
+   * An immutable snapshot of a mediation adapter's initialization status.
    */
-  export type InitializationStatus = {
+  export type AdapterStatus = {
     name: string;
     description: string;
     status: 0 | 1;
-  }[];
+  };
 
   /**
    * The `RequestConfiguration` used when setting global ad settings via `setRequestConfiguration`.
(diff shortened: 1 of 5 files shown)


mikehardy:
haha now the objective-c there is no escaping the lint checks. But you can run them locally for fast cycles...

lib/index.d.ts
```diff
@@ -725,6 +725,17 @@ export namespace GoogleAdsTypes {
...
+  /**
+   * Initialization status of each ad network available to the Google Mobile Ads SDK, keyed by its
+   * GADMAdapter's class name. The list of available ad networks may be incomplete during early
+   * phases of SDK initialization.
+   */
+  export type InitializationStatus = {
+    name: string;
+    description: string;
+    status: 0 | 1;
+  }[];
```
mikehardy, line 737: An interesting choice to make the type itself an array. Typically the type is not the array, and the return is an array of the type ? Seems more ergonomic to me, as then you can address each array entry as the type, as opposed to some part of the type ?

mikehardy: Then the name is maybe not excellent, perhaps GADMAdapterInitializationStatus ? A bit unwieldy, but that captures what it *is* I think?

dylancom: Agree on the ergonomic choice. Will call it "AdapterStatus", that will align with how the types are called in iOS / Android.

mikehardy (approved):
Nice! I like it a lot! Mike+Elliot @ Invertase loved the idea of having you get a real role (literally I quote "the more the merrier" :-) ), so for maintainer purposes my style on these now is to label them "pending merge", which means as soon as CI goes green *anyone* with the capability can merge so long as they pay careful attention to the PR title so it matches conventional commits (and thus drives the next release correctly). I'll add a conventional commits PR title checker action next to make that easier, but that's basically it for merges. Then go to Actions and choose Publish, do a manual run off 'main' and it ...should all work :-)

dylancom:
@mikehardy awesome 🎉! Had no time anymore yesterday evening to try it out, but thanks for the explanation :)

PR #41, opened 2021-12-14
feat!: typescript rewrite

dylancom:
Ready to review


mikehardy:
99 problems but weak typing isn't one

src/types/AdsConsent.interface.ts
```diff
@@ -0,0 +1,365 @@
...
+   *
+   * @param tag The boolean value to tag for under age consent.
+   */
+  setTagForUnderAgeOfConsent(tag: boolean): Promise<void>;
+
+  /**
+   * If using a real device to test, ensure the device ID is provided to the Google Mobile Ads SDK so any mock debug locations
+   * can take effect.
+   *
+   * Emulators are automatically whitelisted and should require no action.
```
mikehardy, line 282:
```suggestion
   * Emulators are automatically on the allowlist and should require no action.
```

google's got a nifty inclusive language filter which makes it trivial to perform what is really one of the more basic gestures of inclusiveness, so I always act on it

src/types/RequestConfiguration.ts
```diff
@@ -0,0 +1,40 @@
...
+   *
+   * You can mark your ad requests to receive treatment for users in the European Economic Area (EEA) under the age of consent.
+   * This feature is designed to help facilitate compliance with the [General Data Protection Regulation (GDPR)](https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX:32016R0679).
+   *
+   * See the [Google Mobile SDK docs](https://developers.google.com/admob/android/targeting#ad_content_filtering) for more information.
+   */
+  tagForUnderAgeOfConsent?: boolean;
+
+  /**
+   * An array of test device IDs to whitelist.
```
mikehardy, line 35:
```suggestion
   * An array of test device IDs to add to the allowlist.
```

src/types/AdEventListener.ts
```diff
@@ -0,0 +1,16 @@
...
+ * A callback interface for all ad events.
+ *
+ * @param type The event type, e.g. `AdEventType.LOADED`.
+ * @param error An optional JavaScript Error containing the error code and message.
+ * @param data Optional data for the event, e.g. reward type and amount
+ */
+export type AdEventListener = (
+  type: AdEventType | RewardedAdEventType,
+  error?: Error,
+  data?: any | RewardedAdReward,
```
mikehardy, line 15: is it really any?

src/types/AdsConsent.interface.ts
```diff
@@ -0,0 +1,365 @@
...
+  /**
+   * A fully formed URL for the privacy policy of the provider.
+   */
+  privacyPolicyUrl: string;
+}
+
+/**
+ * AdsConsentStatus interface.
+ */
+export interface AdsConsentStatus {
```
mikehardy, line 36: weren't these also an enum? the interface and enum feels somehow duplicate to me - are both required?

src/types/AdsConsent.interface.ts
```diff
@@ -0,0 +1,365 @@
...
+   *
+   * // Set outside EEA
+   * await AdsConsentDebugGeography.setDebugGeography(AdsConsentDebugGeography.NOT_EEA);
+   * ```
+   *
+   * @param geography The debug geography location.
+   */
+  setDebugGeography(
+    geography:
+      | AdsConsentDebugGeography['DISABLED']
```
mikehardy, line 205: doesn't indexing in via string defeat the type, why not `AdsConsentDebugGeography.DISABLED` ?

src/types/MobileAdsModule.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * The native module instance for the Google Ads service.
+   */
+  native: any;
+
+  /**
+   * Returns the shared event emitter instance used for all JS event routing.
+   */
+  emitter: any;
```
mikehardy, line 47: would be nice to have a real type

src/types/MobileAdsModule.ts
```diff
@@ -0,0 +1,48 @@
+import { RequestConfiguration } from './RequestConfiguration';
+
+/**
+ * An immutable snapshot of a mediation adapter's initialization status.
+ */
+export type AdapterStatus = {
+  name: string;
+  description: string;
+  status: 0 | 1;
```
mikehardy, line 9: what do the 0 and 1 mean, semantically? Would be nice to have an enum that conveyed semantic meaning in a human way

src/validateAdRequestOptions.ts
```diff
@@ -22,14 +22,14 @@ import {
...
 } from './common';
+import { RequestOptions } from './types/RequestOptions';
 
-export default function validateAdRequestOptions(options) {
-  const out = {};
+export default function validateAdRequestOptions(options?: RequestOptions) {
+  const out: RequestOptions = {};
 
-  if (isUndefined(options)) {
+  if (typeof options === 'undefined') {
```
mikehardy, line 32: is this equivalent to 
```suggestion
  if (options === undefined) {
```
:thinking:

src/validateAdShowOptions.ts
```diff
@@ -15,12 +15,13 @@
...
+import { hasOwnProperty, isBoolean, isObject } from './common';
+import { AdShowOptions } from './types/AdShowOptions';
 
-export default function validateAdShowOptions(options) {
-  const out = {};
+export default function validateAdShowOptions(options?: AdShowOptions) {
+  const out: AdShowOptions = {};
 
-  if (isUndefined(options)) {
+  if (typeof options === 'undefined') {
```
mikehardy, line 24: same === undefined question

mikehardy (changes requested):
First of all, this is amazing. Typescript is clearly the way to go and I am really excited for this.

Second, I have at this point too many decades of experience programming to admit, but I only have a year or so of heavy typescript and no real focus on typing from scratch. Combine the two and sometimes I'm right on the mark with serious issues and sometimes I'm just missing something completely and need an explanation like for a young child :-). Apologies in advance for any young child bits :baby: 

Third, I am of the opinion that in general those doing the work get the last say, and in general you should stay out of their way. Which gives me a strong bias for incremental improvement, light discussion, and merging - relying on the submitter to hold themselves to high standards (which you do, by the way - just stating my style)

So with all that said, I saw some things and I noted some things. And I think with a little discussion they will either be obvious to me, or they will result in improvements, and then without too much fuss we'll merge this and boom, we'll be in typescript.

Thanks!

mikehardy (approved):
Looks great - I don't see any reason to stand in the way

My only tiny thought is I have a slight preference to continue disallowing any and then using a ts-ignore very judiciously when necessary, so that any's don't creep in. And ts-ignores should be in a lint check for eventual fixing + removal. But then you could also say, just have a list of anys (along with ts-ignores) in a lint check for removal and then add the any lint rule back. It's kind of the same, so it's a small thought - and in the end it is not actionable

I'll see if I can get this packed up as a beta release now so it's easy to install it for testing to make sure the packaging is alright

mikehardy:
Well, I got part of what I wanted. We safely have a release that is a major version (v4) and is tagged "beta" on npmjs.com, so "latest" is still the prior v3 version.

That's all good, that's safe.

However, for some reason - and unlike react-native-fbsdk-next when I did it, it got the npmjs.com release label right but it did not tag the release version itself in package.json as '-beta1'. Why? I don't know. But version 4.0.0 is "burnt" / used, so whenever this actually releases - if we need changes - it will be as 4.0.1 or something.

Not a big deal but it's odd, and I wish it had -beta1 in the version name. But I can't figure out why it didn't and in general I don't like beta releases, I prefer testing and then just "release" releases ;-), so I gave it a short timebox to figure it out and came empty and I'm leaving it for now

It's out! :-)

mikehardy:
@Mikenso + @birdofpreyru if you have a chance, your help testing v4.0.0 would be appreciated. It is currently on the `@beta` label in npmjs.com, so you have to install it specifically via version or via beta tag. The goal is to release this in a week or so if it seems fine, but it would be great to hear real success reports (or failure reports!). Thanks for your past help here and thanks in advance if you get a chance to check v4. Cheers

birdofpreyru:
Hey @mikehardy , I am a bit busy with different stuff now. Time allowing, I'll probably give it a try second half ot this week.

Mikenso:
Hi @mikehardy, I would like to test as soon as I will have free time. 
But from quick look - It is pleasure to see **dylancom** got rid from many default exports, I never understood people's love for them

Mikenso:
I have tested interstitial and banner at my current project written with ts.
And also checked at just created for test purposes raw js project.
It seems to work as it were at current stable version.

mikehardy:
Fantastic! Thanks very much for the feedback @Mikenso

PR #49, opened 2021-12-29
feat: extended error logging

dylancom:
### Description

Some errors remain "unknown" and therefore in some cases it's hard to understand what goes wrong in case of errors.


mikehardy (approved):
This looks great to me - my only thought is if there is still some sort of unknown, is there any extra information we may print about it at all? Best to chuck everything we have into a log or something if possible, for best self-service. If there is nothing else to print there is nothing else to do though :-), feel free to merge+release as it looks good to me

PR #50, opened 2021-12-29
feat: smarter error logging

dylancom:
### Description

This makes use of the messages that the google mobile ads sdk already provides on error objects.
So there is no need anymore to provide custom messages.


ios/RNGoogleAds/RNGoogleAdsCommon.m
```diff
@@ -87,55 +87,38 @@ + (GADRequest *)buildAdRequest:(NSDictionary *)adRequestOptions {
 
 + (NSDictionary *)getCodeAndMessageFromAdError:(NSError *)error {
```
mikehardy, line 88: I notice in Java you changed the method signature but not in Obj-C - I suppose it's best to have them both line up?

mikehardy (changes requested):
I like it - one tiny comment about API harmony across the platforms, but it's an internal API so that's minor. The real purpose of the change I like a lot

mikehardy (approved):
LGTM! I would go with squash merge for a clean commit history on main branch, once CI passes

PR #53, opened 2022-01-20
refactor: use type predicates

dylancom:
Currently some type validation helpers from common weren't used as those resulted in TS errors.
By defining a "Type predicate", TS understands the validation functions.
https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates


mikehardy (approved):
This is cool! Nice

PR #63, opened 2022-01-26
chore: merge beta branch to main branch

mikehardy:
Should be a rebase merge once beta branch has successful integration reports (@dylancom ? @birdofpreyru ? )

This should be the last large-scale change for the module I hope - it's been successfully extracted (v1), handled some API updates (v2, v3) and now has it's main maintainership hurdles cleared with typescript and a final name (v4)... hopefully not famous last words but I'm hoping it settles down now and we focus on finding + eliminating all errors so it's rock stable now


birdofpreyru:
@mikehardy 4.0.2 seems to work fine for me on Android, not that I am doing any in-depth testing, nor have arrived to try in on iOS yet.

I also just noticed that you changed the package name, I think it would be nice to add to https://www.npmjs.com/package/@invertase/react-native-google-ads a note about the new name, and also push there a fake major version, so that anybody depending on that, but not following the github closely has a better chance to notice the re-naming.

Also, not sure the new name **react-native-google-mobile-ads** makes it easy discoverable, my guess is people looking for such RN library would google "React Native Admob", and will find one of many other libs with name similar to **react-native-admob**, e.g. [@react-native-admob/admob](https://www.npmjs.com/package/@react-native-admob/admob). I would go with **@invertase/react-native-admob**.

mikehardy:
Thanks for testing 4.0.2 which contains the typescript rewrite, that was the big change. The name change was actually not that disruptive and included a lot more testing (https://github.com/invertase/react-native-google-mobile-ads/blob/c2b109d30f694a9a710d71a8c80b29d955caffb2/example/App.tsx#L186-L189) so I'm pretty confident in that one

Great point about posting a note / package to the old name - I'll do that

There are only two things hard in computing: caching and naming things. I think you have a valid point about a possible different name but when we considered all possible naming permutations (each one seems to already have a package in existence!) we went with this one. It contains "react-native" which is important", and "google mobile ads" is the actual name of the underlying SDK, it should be okay. We'll have to work on discoverability with package.json keywords and just simply being a better package over time...

birdofpreyru:
@mikehardy **react-native-google-mobile-ads@4.1.0** fails to build within a host project on Android:
image
I guess, it is a one-liner fix, but I am Gradle-noob, so hints how to correct it are welcome :)

mikehardy:
I think that's a change needed in app.json - should not need a PR, should need a rename of the element in app.json? Or it could be something I missed. All I can tell you is I *swear* I checked android builds and had the example app running and showing android ads up to the final commit (for app open ads). I did run into this myself though, at one point during the rename. Either way, looking forward to seeing the PR and I will test locally

Sorry for the bother

birdofpreyru:
Ogh, indeed I did not rename the key in `app.json`.
I tried to change `android/build.gradle` this way:
```diff
-if (rootProject.ext.googleMobileAdsJson) {
+if (rootProject.ext.hasProperty("googleMobileAdsJson")) {
```
it helped to pass the build, but then the app started to crash on the start-up with a segmentation fault and no clear error message. I guess I just need to change `app.json`, and no need to modify the library code. Will try now.

mikehardy:
I am going to guess that works, and that will scream for a callout in the changelog, which I will add immediately pending your test result

birdofpreyru:
@mikehardy I am confirming, 4.1.0 works (builds and does not crash), my bad not updating `app.json`.

However, now I am getting this runtime error:
> _Cannot read property 'requestInfoUpdate' of null_

Did you remove `AdsConsent` export already?

mikehardy:
:thinking: there should not have been a change related to that. This version was 1) typescript, 2) module name 3) addition of app open

There was not the intention to regress anywhere. I'll investigate - if you can provide details, anything might help locate it and/or add tests to the example app so it is conclusively working

Thanks again for testing, now that I have confirmation on the app.json thing I'll update the changelog. Oversight on my part, that's obviously important to mention

birdofpreyru:
I don't know... I experience something really weird I never saw before: I print `AdsConsent` to console and verify that it is an object, and it has `.requestInfoUpdate()` method. On the very next line I call it, and get the error saying I'm trying to call a method of `null`. Just reverting back to 4.0.2, everything works fine. I guess, there should be something peculiar or wrong with my app setup (just JS by the way) conflicting with some TypeScript changes, but I guess I'll stick with 4.0.2 for now, and come back to trying 4.1.0 later :)

mikehardy:
Okay - that's still a great report, I really appreciate it. The test cases right now *do* show the 4 types of ads working but do not probe AdConsent at all. And my family is really tired of me hearing it by now but "if it isn't tested, it's broken...", so I'm going to assume we broke something somehow, and I'll add a test. Cheers

dylancom:
@birdofpreyru Thanks again for the report and testing so quickly! I just published a fix.

PR #76, opened 2022-02-08
feat: custom targeting

frankenthumbs:
### Description

We want to move onto this library after using a fork of `https://github.com/nfl/react-native-admob` named `https://github.com/nfl/react-native-admob`. We have found that customTargeting values and the ability to refresh banner ads was not yet available to adding that capability here.

rebuilt changes to get rid of some git history from this PR: https://github.com/invertase/react-native-google-mobile-ads/pull/72

### Release Summary

- customTargeting in `requestOptions`
- requestAd function in `GoogleMobileAdsBannerView`

### Checklist

  - [ ] Yes
  - [X] `Android`
  - [X] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

### Test Plan

Not sure what e2e/jest tests I have to add. If i can get some direction there. please and thank you 🙏 

:fire:

---


android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsBannerAdViewManager.java
```diff
@@ -56,6 +57,16 @@ public String getName() {
...
   }
 
+  @Override
+  public void receiveCommand(final ReactViewGroup root, String commandId, ReadableArray args) {
+    // This will be called whenever a command is sent from react-native.
+    switch (commandId) {
+      case "requestAd":
+        requestAd(root);
+        break;
+    }
```
mikehardy, line 67: do you think it is worthwhile to include a default here and Log.d what the command was for developer usage? I'm not familiar with the AdMob receiveCommand use case specifically, so that may be useless, but adding developer logging and including a default branch is a typical linter suggestion, so I thought I'd mention it

mikehardy:
Thanks again for proposing this, and for the extra effort to make a clean commit trail.
- It looks like a local run of `lint:android` might be needed until it runs clean (you can just keep re-running it) then commit the result
- I had a worry about nullability, and a question about logging

Other than that, it appeared like it would work and work well, but Dylan probably has some thoughts since he has more experience with AdMob internals

src/TestIds.ts
```diff
@@ -20,6 +20,7 @@ import { Platform } from 'react-native';
 export const TestIds = {
   APP_OPEN: '',
   BANNER: '',
+  CUSTOM_TARGETING: {},
```
dylancom, line 23: TestIds serves as an object that holds al the Test Ad Unit ID's -> https://developers.google.com/admob/ios/test-ads
Whereas customTargeting is a request option.

dylancom:
I'm not sure why we would want to add a requestAd method though. The BannerAd refreshes itself and you can adjust the rate within the AdMob dashboard.

dylancom:
customTargeting seems to be missing in `ReactNativeGoogleMobileAdsCommon.java`

frankenthumbs:
> I'm not sure why we would want to add a requestAd function though. The BannerAd refreshes itself and you can adjust the rate within the AdMob dashboard.
we use it (infrequently) when certain customTargeting params are changed and adunitId changes.. an instance of the first use case is when we have a "companion" ad next to a video preroll, they have to sync up.

> customTargeting seems to be missing in `ReactNativeGoogleMobileAdsCommon.java`
ah thank you! when i read the android code I thought it would all just work. still very much learning android.

> * It looks like a local run of `lint:android` might be needed until it runs clean (you can just keep re-running it) then commit the result
tried running `lint:android` and didn't get any errors.. will keep an eye out on the new tests

> * I had a worry about nullability, and a question about logging
hopefully the update fixed the nullability and logging concern.

example/App.tsx
```diff
@@ -119,6 +119,7 @@ class BannerTest implements Test {
           unitId={bannerAdUnitId}
           size={BannerAdSize.ADAPTIVE_BANNER}
           requestOptions={{
+            customTargeting: {},
```
dylancom, line 122: If we just pass in an empty obj, why do we use it at all in the example?

dylancom:
Sorry for the inconvenience, but I prefer to have a separate PR for the banner reloading, so we can test this in isolation. Ideally we would also have a test for it in the example app with a button stating for e.g. "(Not recommended) manual refresh)".

frankenthumbs:
> Sorry for the inconvenience, but I prefer to have a separate PR for the banner reloading, so we can test this in isolation. Ideally we would also have a test for it in the example app with a button stating for e.g. "(Not recommended) manual refresh)".

that's fair.. i'm also not against keeping our own fork if this doesn't sound appealing to put in y'alls repo.

dylancom:
> > Sorry for the inconvenience, but I prefer to have a separate PR for the banner reloading, so we can test this in isolation. Ideally we would also have a test for it in the example app with a button stating for e.g. "(Not recommended) manual refresh)".
> 
> that's fair.. i'm also not against keeping our own fork if this doesn't sound appealing to put in y'alls repo.

Why not just set a unique key on BannerAd?
`<BannerAd
        key={bannerKey}
        unitId={bannerAdUnitId}
        size={BannerAdSize.ADAPTIVE_BANNER}
      />`
      
If you then update the key prop, a new banner will be rendered.

frankenthumbs:
> > > Sorry for the inconvenience, but I prefer to have a separate PR for the banner reloading, so we can test this in isolation. Ideally we would also have a test for it in the example app with a button stating for e.g. "(Not recommended) manual refresh)".
> > 
> > 
> > that's fair.. i'm also not against keeping our own fork if this doesn't sound appealing to put in y'alls repo.
> 
> Why not just set a unique key on BannerAd? `<BannerAd key={bannerKey} unitId={bannerAdUnitId} size={BannerAdSize.ADAPTIVE_BANNER} />`
> 
> If you then update the key prop, a new banner will be rendered.

thats smart! didn't think about using the key prop that way!

mikehardy (approved):
looks really clean to me, for what it's worth, thanks for iterating on it @frankenthumbs

frankenthumbs:
`yarn lint:code` runs locally. not sure if random CI issue or environment differences.

mikehardy:
Strange - I've had issues with that check before ([I'm actually the maintainer there](https://github.com/invertase/nodejs-google-java-format/commits/main) :sweat_smile: ) but it has always been local that fails, CI has never given me flaky signals on it.

I re-ran the check and it failed again so there must be something to it

I checked it locally and indeed the java imports were not correct? They just needed a quick re-ordering. I committed that and pushed the commit to the branch so it should be good now

dylancom (approved):

PR #85, opened 2022-03-02
test(example): add AdConsent test

### LINKED ISSUE #64: Example app needs demonstration of consent management
Looks like example/App.tsx could use a probe of AdConsent vs current documentation to make sure that it is working, as it seems to be broken: 
----

I don't know... I experience something really weird I never saw before: I print `AdsConsent` to console and verify that it is an object, and it has `.requestInfoUpdate()` method. On the very next line I call it, and get the error saying I'm trying to call a method of `null`. Just reverting back to 4.0.2, everything works fine. I guess, there should be something peculiar or wrong with my app setup (just JS by the way) conflicting with some TypeScript changes, but I guess I'll stick with 4.0.2 for now, and come back to trying 4.1.0 later :)

_Originally posted by @birdofpreyru in https://github.com/invertase/react-native-google-mobile-ads/issues/63#issuecomment-1022556563_

birgernass:
### Description

Add a test case for the AdConsent interface to `App.tsx`.
The test case does not work with the default AdMob App IDs.
The UMP has to be configured in the AdMob/Ad Manager backend for the test case to work.

### Related issues

Closes #64 

---

### DIFF
--- a/example/App.tsx
+++ b/example/App.tsx
@@ -1,8 +1,18 @@
 import React from 'react';
-import {Button, SafeAreaView, ScrollView, StyleSheet, View} from 'react-native';
+import {
+  Button,
+  SafeAreaView,
+  ScrollView,
+  StyleSheet,
+  Text,
+  View,
+} from 'react-native';
 import {Test, TestRegistry, TestResult, TestRunner, TestType} from 'jet';
 
 import {
+  AdsConsent,
+  AdsConsentDebugGeography,
+  AdsConsentStatus,
   AppOpenAd,
   InterstitialAd,
   TestIds,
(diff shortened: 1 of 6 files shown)


example/App.tsx
```diff
@@ -182,11 +185,55 @@ class RewaredTest implements Test {
...
+  }
+
+  getTestType(): TestType {
+    return TestType.Interactive;
+  }
+
+  render(onMount: (component: any) => void): React.ReactNode {
+    return (
+      <View style={styles.testSpacing} ref={onMount}>
+        <Button
```
mikehardy, line 200: A shame about the constraint that it just won't work with the test ad ids (seems like an oversight really, you can't fully exercise something really important!) but I'm okay with a test existing as long as it gets something like 
```suggestion
        <Text>This will not work with the test App ID. For this to work you must configure your real App ID and The UMP has to be configured in the AdMob/Ad Manager backend for the test case to work.</Text>
        <Button
```

mikehardy: I guess that'll require the Text import up top as well

mikehardy (changes requested):
Can you confirm that with your App ID configured it worked for you in that test case?
If so, then basically just adding a disclaimer to the test for now works for me, and this is fantastic!

Related thought - Invertase has a firebase project that is maintained solely for testing react-native-firebase. What do you think the worst-case scenario is if we create an AdMob account just so we can have some real IDs and configured UMP stuff in order to make this testable / runnable out of the box? I can't imagine a real abuse scenario - who would want to use someone else's app id for ads and give up revenue? - but keen to hear wise opinions on why this is a bad idea

mikehardy:
I'm sure your answer will be "yes I tested it with a real app id and it worked" by the way, so just wanted to add that I'll do a release with the new consent stuff as soon as I hear that yes, I just wanted to make sure everyone had their chance to try it, and that there was some confirmation it worked before release

birgernass:
Yes, tested it and it works.

I don't know anything about the potential downsides of such a dummy account I'm afraid. I like the idea though.

example/App.tsx
```diff
@@ -182,11 +192,64 @@ class RewaredTest implements Test {
...
+              consentInfo.isConsentFormAvailable &&
+              consentInfo.status === AdsConsentStatus.REQUIRED
+            ) {
+              await AdsConsent.showForm();
+            }
+          }}
+        />
+
+        <Text>
+          This test case will not work with the test App ID. You must configure
```
mikehardy, line 225: even better - the specifics here will be a big help I'm sure, thanks for taking the time to put all the details in

mikehardy (approved):
looks fantastic, okay :crossed_fingers:  for the release now (the release scripts have been working well but I'm always surprised by the way releases can fail)

Issue #96, opened 2022-03-16
Ads consent with the new UMP SDK implementation


birdofpreyru:
Sorry for not following the issue template, I am just trying to migrate my projects to v5 of this library, and need some place to record issues I encounter on the way :)

- Docs: [European User Consent](https://docs.page/invertase/react-native-google-mobile-ads/european-user-consent) page should mention that configuration is required in AdMob web console, at https://apps.admob.com/v2/privacymessaging. Right now it can be figured out from the link at [Installation](https://docs.page/invertase/react-native-google-mobile-ads#european-user-consent) page, but not on EUC page itself. Also, in my current understanding, it is not necessary for the library user to read entire [Obtaining Consent with the User Messaging Platform](https://developers.google.com/admob/ump/android/quick-start) document, for purposes of setting up.

- I am suspecting a bug in the library. I am creating GDPR form with three options (Consent / Do not consent / Manage options). It seems to work as expected when I do consent (including banner / interstitial ads being shown as expected, at the first glance). However, when I choose to **not consent** in the form, I am still getting `{"status": "OBTAINED"}` result from the **.showForm()** call (although no ads are served in that case).

birgernass:
Not receiving ads with the consent status being obtained is not a bug. You can find a bit more information about the change and how to get the consent information on the [migration page](https://docs.page/invertase/react-native-google-mobile-ads/migrating-to-v5) of the docs.

birdofpreyru:
@birgernass My main point was that I was getting `{"status": "OBTAINED"}` from **.showForm()** when I declined the consent in the form. I see, in the migration page it mentions that this status in this case just means the user have seen the form already and it is necessary to

> You can read the IABTCF_TCString key from standardUserDefaults / SharedPreferences and decode it with a library like [@iabtcf/core](https://github.com/InteractiveAdvertisingBureau/iabtcf-es/tree/master/modules/core#iabtcfcore) though.

It is however different from what is written in the main docs (https://docs.page/invertase/react-native-google-mobile-ads/european-user-consent#gathering-user-consent), and also IMHO counter-intuitive... why not return `{"status": "REQUIRED"}` if the consent requried, but user denied it in the form?

birgernass:
The interfaces tries to stick as close to the native implementation as possible. Required means that the user has not yet seen the modal and that you should show it.

birdofpreyru:
[...]

PR #103, opened 2022-03-28
feat(consent): add method that returns consent choices

birgernass:
### Description

v5.0.0 introduced the UMP SDK, which stores the consent as an encoded string on the device and doesn't provide further information. It is a common use case to act upon a user's choices though, especially as the Google Mobile Ads SDK won't serve any ads if the user doesn't at least provide consent to store and/or access information on the device. This PR aims to bridge that gap by introducing a new method to the `AdsConsent` interface called `getUserChoices` (happy to hear other suggestions).

I also went ahead and replaced the JSDocs inside `src/AdsConsent.ts` with TypeScript types. If the JSDocs serve a purpose, then let me know and I'll revert.

### Related issues

Related to https://github.com/invertase/react-native-google-mobile-ads/issues/96

### Checklist

  - [X] `Android`
  - [X] `iOS`
- My change includes tests;
  - [X] `e2e` tests added or updated in `__tests__e2e__`
- This is a breaking change;
  - [X] No

### DIFF
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsConsentModule.java
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsConsentModule.java
@@ -16,11 +16,13 @@ package io.invertase.googlemobileads;
  * limitations under the License.
  *
  */
-
+import android.content.SharedPreferences;
+import android.preference.PreferenceManager;
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactMethod;
+import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
 import com.google.android.ump.ConsentDebugSettings;
@@ -28,7 +30,6 @@ import com.google.android.ump.ConsentInformation;
 import com.google.android.ump.ConsentRequestParameters;
 import com.google.android.ump.UserMessagingPlatform;
 import io.invertase.googlemobileads.common.ReactNativeModule;
-import java.util.List;
 import javax.annotation.Nonnull;
 
 public class ReactNativeGoogleMobileAdsConsentModule extends ReactNativeModule {
@@ -63,10 +64,10 @@ public class ReactNativeGoogleMobileAdsConsentModule extends ReactNativeModule {
           new ConsentDebugSettings.Builder(getApplicationContext());
 
       if (options.hasKey("testDeviceIdentifiers")) {
-        List<Object> devices = options.getArray("testDeviceIdentifiers").toArrayList();
+        ReadableArray devices = options.getArray("testDeviceIdentifiers");
 
-        for (Object device : devices) {
-          debugSettingsBuilder.addTestDeviceHashedId((String) device);
+        for (int i = 0; i < devices.size(); i++) {
+          debugSettingsBuilder.addTestDeviceHashedId(devices.getString(i));
         }
       }
 
@@ -151,5 +152,18 @@ public class ReactNativeGoogleMobileAdsConsentModule extends ReactNativeModule {
   @ReactMethod
   public void reset() {
     consentInformation.reset();
+  }
+
+  @ReactMethod
+  public void getTCString(Promise promise) {
+    try {
+      SharedPreferences prefs =
+          PreferenceManager.getDefaultSharedPreferences(getReactApplicationContext());
+      // https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#in-app-details
+      String tcString = prefs.getString("IABTCF_TCString", null);
+      promise.resolve(tcString);
+    } catch (Exception e) {
+      rejectPromiseWithCodeAndMessage(promise, "consent-string-error", e.toString());
+    }
   }
 }
--- a/docs/common-reasons-for-ads-not-showing.mdx
+++ b/docs/common-reasons-for-ads-not-showing.mdx
@@ -13,6 +13,7 @@ You'll be notified of approval or rejection via email after review.
 
 New apps and ad units take some time to activate.
 Here are common reasons you may not see live impressions immediately:
+
 - It usually takes at least an hour after you create an app or ad unit
 - Sometimes it can take a few days for ads to appear in new apps or ad units
 - New iOS apps will not show Google ads until they’re listed in the Apple App Store
@@ -35,6 +36,7 @@ Test devices can either be added in the AdMob UI or programmatically using the G
 Follow the steps below to add your device as a test device.
 
 #### Add your test device in the AdMob UI
+
 For a simple, non-programmatic way to add a test device and test new or existing app builds, use the AdMob UI.
 [Learn how](https://support.google.com/admob/answer/9691433).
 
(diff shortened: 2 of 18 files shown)


mikehardy:
This took a few conflicts from recent PR merges, but I want it to go in vs succumb to entropy so I since I was the one merging and blowing up PRs, I pulled this locally, rebased it to main, fixed the conflicts, then tested it out thorougly. I can say it has no regressions on android or ios emulators

Still needs a review but testing is half the battle I think - takes as much time as the review normally :-)

mikehardy:
By the way - the consent form seems to come up correctly on my iOS simulator, which is kind of nice. Not on the Android Emulator, but seeing it pop up was cool. I had never even tried it believe it or not, because of the ominous warning in there. Do you think there is any way to expose the decode of the consent choices as a console.log at least?

src/AdsConsentPurposes.ts
```diff
@@ -0,0 +1,175 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *
+ */
+
+export enum AdsConsentPurposes {
```
mikehardy, line 19: Did these come from here somewhere? https://github.com/InteractiveAdvertisingBureau/iabtcf-es/tree/master/modules/core

It would be good to have a reference  in the comments of how these came to be as I imagine they will change over time

src/AdsConsentSpecialFeatures.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *
+ */
+
+export enum AdsConsentSpecialFeatures {
```
mikehardy, line 19: Same as the Features one - would be great to have a comment pointing to the source of these or the mechanism for coming up with them, for future maintenance

src/types/AdsConsent.interface.ts
```diff
@@ -127,3 +168,175 @@ export interface AdsConsentInfo {
    */
   isConsentFormAvailable: boolean;
 }
+
+/**
+ * The options used when requesting consent information.
+ */
+export interface AdsConsentUserChoices {
```
mikehardy, line 175: Same concern as the other two spots, this seems like an area that will need semi-frequent refresh from source to get the consent correct as it changes over time. This one actually looks pretty duplicate too - is there a way to de-duplicate these three files so the same text is only in one spot? There may not be but it looks like a don't-repeat-yourself violation

ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
```diff
@@ -121,4 +121,17 @@ - (NSString *)getConsentStatusString:(UMPConsentStatus)consentStatus {
 
 RCT_EXPORT_METHOD(reset) { [UMPConsentInformation.sharedInstance reset]; }
 
+RCT_EXPORT_METHOD(getTCString : (RCTPromiseResolveBlock)resolve : (RCTPromiseRejectBlock)reject) {
+  @try {
+    NSString *tcString = [[NSUserDefaults standardUserDefaults] objectForKey:@"IABTCF_TCString"];
```
mikehardy, line 126: Is this key a constant from the form itself (here and in Java), or some other spot? I wonder if the key could be imported as a symbol so it wasn't just a magic string

More importantly, if we're storing and decoding persistent state on disk, is there any provision for versioning, checking for staleness, updating contents of that string if the form has changed? This may all be a function of the library, I am very new to this domain

mikehardy (changes requested):
Mostly nit-picks and questions I don't see anything fundamentally wrong here, and at though there is no place in the example app where this is actually exercised, it is at least not a regression at the moment - all current functionality works. Should switch to approve quickly after short discussion

mikehardy:
Great thanks! I just realized I should have issued a little bit of a warning: since I pushed to your branch on your fork, please make sure you git rebase from it before you start working locally again or things could get messy. If you've already done that or knew that sorry for boring you but if you haven't, or it gets weird and you need a hand just let me know :-), my goal was to be helpful and speed things up vs making it worse, hopefully that's the case 🤞

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsConsentModule.java
```diff
@@ -63,10 +64,10 @@ public void requestInfoUpdate(@Nonnull final ReadableMap options, final Promise
           new ConsentDebugSettings.Builder(getApplicationContext());
 
       if (options.hasKey("testDeviceIdentifiers")) {
-        List<Object> devices = options.getArray("testDeviceIdentifiers").toArrayList();
+        ReadableArray devices = options.getArray("testDeviceIdentifiers");
 
-        for (Object device : devices) {
-          debugSettingsBuilder.addTestDeviceHashedId((String) device);
+        for (int i = 0; i < devices.size(); i++) {
+          debugSettingsBuilder.addTestDeviceHashedId(devices.getString(i));
```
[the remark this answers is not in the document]
birgernass: When I tried react-native nightly a few weeks ago, playing around with the new architecture, I got build issues due to these typings. I didn't dig deep enough to tell what the reason was, but I figured it wouldn't hurt to change.

docs/common-reasons-for-ads-not-showing.mdx
```diff
@@ -48,39 +50,39 @@ If you want to test ads in your app as you're developing, follow the steps below
...
-    ```
+   mobileAds()
+     .setRequestConfiguration({
+       // An array of test device IDs to add to the allow list.
+       testDeviceIdentifiers: ['2077ef9a63d2b398840261c8221a0c9b', 'EMULATOR'],
+     })
+     .then(() => {
+       // Request config successfully set!
+     });
+   ```
```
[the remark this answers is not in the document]
birgernass: Yes, I agree, I rebased and split the commit into two.

src/AdsConsent.ts
```diff
@@ -75,18 +76,67 @@ export const AdsConsent: AdsConsentInterface = {
...
+
+  async getUserChoices(): Promise<AdsConsentUserChoices> {
+    const tcString = await native.getTCString();
+
+    let tcModel: TCModel;
+
+    try {
+      tcModel = TCString.decode(tcString);
+    } catch (e) {
+      tcModel = new TCModel();
```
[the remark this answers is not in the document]
birgernass: I added such a log.

src/AdsConsentPurposes.ts
birgernass: I added links and a very brief description, let me know if that is not what you had in mind.

src/types/AdsConsent.interface.ts
birgernass: Im not sure how this works with TypeScript, editor resolution and the like. Wanted to have the description showing up during code completion as well as next to their definitions. Do you have a goof idea? Happy to change.

ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
birgernass: I sadly couldn't find an export. They refer to using this string inside their documentation as well, I referenced it in the code comments.

We do not interact with the TCString besides reading it from disk. Google and 3rd party vendors use the content of the string to check for available consent and other information. The UMP SDK stores it on the device and also automatically handles outdated values, e.g. if Google itself adds new vendors to their list. The consent modal will again be shown to the user.

The IAB and the Consent Framework are slow moving industry standards, generally speaking, and the key should stay relevant for quite a while.

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsConsentModule.java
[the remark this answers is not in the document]
mikehardy: That's as solid a defense as any. I know they are switching up some of the marshal/unmarshal types, could be an artifact of that. Anyway, works for me!

ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
mikehardy: Okay, that makes sense - what I understood was "the libraries handle this", and the format does seem forward-compatible, sort of JSON-y

src/AdsConsentPurposes.ts
mikehardy: perfect thanks - the added links sprinkled throughout at least give a handle to all the context an interested dev (like myself) needs to go digging.

src/types/AdsConsent.interface.ts
mikehardy: No better ideas if you don't have any. Better too much doc than too little, no reason to hold up on this

mikehardy (approved):
LGTM - everything looks completely addressed, by me read. Thanks a bunch for this, it's on "release next thing" track in my  mind

PR #108, opened 2022-04-08
test: test extensions / enhancements

mikehardy:
Nothing in here is functional in the library, it is just extra test code to exercise more of the hooks + hook features and add lots of state logging so an example user can see clearly what is going on

I used this to exercise the hooks code in PR #100


mikehardy:
It's not great form for a maintainer to just merge their code - we need review too! - but this is literally nothing but test code, and mostly just log statements on top of that. So I'll merge it pending CI

mikehardy:
Took the opportunity to run `refresh-example.sh` to get a fresh RN68 example and then re-verify everything with the new logging.

Rewarded Ads never seem to fill on iOS (they never have?), but everything else works on both platforms.

mikehardy:
I asked the SDK forum about iOS rewarded ad no-fills: https://groups.google.com/g/google-admob-ads-sdk/c/5ixRSHOkhHE

PR #115, opened 2022-04-19
feat: migrate to sdk v9 (ios), upgrade to sdk 20.6.0 (android)

### LINKED ISSUE #113: Update to v9 of Google Ads iOS SDK
the v9 series of google's ios mobile SDK is out and has a few breaking changes to handle

The main one appears to be that no ads will serve on iOS10 any longer, though we still have iOS10 set as our minimum target version at the moment. So that will need a CHANGELOG / BREAKING CHANGE note

The other breaking changes *may* all be internal and not actually break module consumers here if they do not affect our APIs? That's unknown

@dylancom mentions he's played with this already and it wasn't bad?

https://developers.google.com/admob/ios/rel-notes
https://developers.google.com/admob/ios/migration

Could stack this with contemplated breaking change for event listener APIs from @wjaykim

### LINKED ISSUE #114: Update to v21 APIs for Google Mobile Ads Android SDK
No investigation has been done yet to see how the API changes affect our usage or module user's usage transitively - it's a breaking change for the underlying SDK but may or not be a breaking change for module users

Related #113 - and changes could probably be batched together with that one either way (breaking them together if it's a breaking change, with v6 release of this module) or releasing them together but not breaking

Note: v21 is not released quite yet but likely will be shortly
https://developers.google.com/admob/android/migration
https://developers.google.com/admob/android/rel-notes

dylancom:
### Description

Fixes #113 / Fixes #114

**Major upgrade to 9.3.0 on iOS. (See: https://developers.google.com/admob/ios/migration)**
- Rename adDidPresentFullScreenContent: to adWillPresentFullScreenContent:
There is no behavior change. The delegate method is invoked right before the ad is about to be presented, so the new method name better reflects its functionality.

- Remove location setting API on GADRequest
(void)setLocationWithLatitude:longitude:accuracy: has been deleted from GADRequest since location data is not used by Google to target ads. Use third-party APIs to provide the information to third-party ad networks if required.

**Minor upgrade from 20.5.0 to 20.6.0 on Android.**


mikehardy:
Screwed up my first attempt to chunk these up into separate units, but I think I got it correct the second time.
This is great - really not much, as you said, to adapt. in fact by my read the Android one is already going to work without additional changes now that the location stuff is gone, to do iOS v9

Mostly the commit adds the breaking change notice that I want for the merge, I will wait to release it until we are done with the listener + GAM break and get one v6 out with all of them

mikehardy (approved):
a thing of beauty

PR #119, opened 2022-04-26
refactor: move common ad logics to mobilead class

wjaykim:
### Description

Moved common `load` and `show` methods in each full screen ad classes into `MobileAd` class.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsAppOpenModule.java
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsAppOpenModule.java
@@ -119,7 +119,8 @@ public class ReactNativeGoogleMobileAdsAppOpenModule extends ReactNativeModule {
   }
 
   @ReactMethod
-  public void appOpenShow(int requestId, ReadableMap showOptions, Promise promise) {
+  public void appOpenShow(
+      int requestId, String adUnitId, ReadableMap showOptions, Promise promise) {
     if (getCurrentActivity() == null) {
       rejectPromiseWithCodeAndMessage(
           promise,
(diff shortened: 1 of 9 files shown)


android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsAppOpenModule.java
```diff
@@ -119,7 +119,8 @@ public void onAdFailedToLoad(@NonNull LoadAdError loadAdError) {
   }
 
   @ReactMethod
-  public void appOpenShow(int requestId, ReadableMap showOptions, Promise promise) {
+  public void appOpenShow(
+      int requestId, String adUnitId, ReadableMap showOptions, Promise promise) {
```
mikehardy, line 123: That's odd :thinking: - there is no action necessary, but if I understand correctly this is a spacing-only change? And I thought we were enforcing spacing/formatting with google-java-format via https://github.com/invertase/react-native-google-mobile-ads/blob/8bbc0ad742bb0a23b6fecdd718e701985d21086b/package.json#L71

So I wouldn't expect to see any changes like this unless lint demanded them. So either it was failing before and CI has a false-positive (so needs investigation) *or* google-java-format allows both? *or* something unknown.

Not a big deal at all but I try to use auto-formatting and auto-format-checks as much as possible so that diffs are small and no one argues about spacing (total waste of time in my opinion haha)

mikehardy (approved):
That's a nice refactor - this looks great and the tests appear to work well when I pulled the PR and tested

PR #129, opened 2022-05-02
feat: add destroy method

wjaykim:
### Description

Adds `destroy` method for memory cleanup. Developers can call the method in appropriate timing.
Also, ad instance in native side code will be automatically removed from reference array after the ad is dismissed.
This changes may reduce some memory leaks.

### Release Summary

Adds `destroy` method for full screen ads.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/ios/RNGoogleMobileAds/RNGoogleMobileAdsAppOpenModule.h
+++ b/ios/RNGoogleMobileAds/RNGoogleMobileAdsAppOpenModule.h
@@ -23,7 +23,6 @@
 
 @interface RNGoogleMobileAdsAppOpenModule : NSObject <RCTBridgeModule>
 
-@property(strong, nonatomic) GADAppOpenAd* appOpenAd;
-@property(strong, nonatomic) RNGoogleMobileAdsFullScreenContentDelegate* appOpenDelegate;
++ (void)removeAppOpen;
 
 @end
--- a/ios/RNGoogleMobileAds/RNGoogleMobileAdsAppOpenModule.m
+++ b/ios/RNGoogleMobileAds/RNGoogleMobileAdsAppOpenModule.m
@@ -21,6 +21,9 @@
 #import "RNGoogleMobileAdsCommon.h"
 #import "common/RNSharedUtils.h"
 
+static __strong GADAppOpenAd *appOpen;
+static __strong RNGoogleMobileAdsFullScreenContentDelegate *appOpenDelegate;
+
 @implementation RNGoogleMobileAdsAppOpenModule
 #pragma mark -
 #pragma mark Module Setup
@@ -42,7 +45,6 @@ RCT_EXPORT_METHOD(appOpenLoad
                   : (nonnull NSNumber *)requestId
                   : (NSString *)adUnitId
                   : (NSDictionary *)adRequestOptions) {
-  self.appOpenAd = nil;
   [GADAppOpenAd loadWithAdUnitID:adUnitId
                          request:[RNGoogleMobileAdsCommon buildAdRequest:adRequestOptions]
                      orientation:UIInterfaceOrientationPortrait
@@ -58,14 +60,14 @@ RCT_EXPORT_METHOD(appOpenLoad
                                                    data:nil];
                    return;
                  }
-                 self.appOpenAd = appOpenAd;
+                 appOpen = appOpenAd;
                  RNGoogleMobileAdsFullScreenContentDelegate *fullScreenContentDelegate =
                      [[RNGoogleMobileAdsFullScreenContentDelegate alloc] init];
                  fullScreenContentDelegate.sendAdEvent = GOOGLE_MOBILE_ADS_EVENT_APP_OPEN;
                  fullScreenContentDelegate.requestId = requestId;
                  fullScreenContentDelegate.adUnitId = adUnitId;
-                 self.appOpenAd.fullScreenContentDelegate = fullScreenContentDelegate;
-                 self.appOpenDelegate = fullScreenContentDelegate;
+                 appOpen.fullScreenContentDelegate = fullScreenContentDelegate;
+                 appOpenDelegate = fullScreenContentDelegate;
                  [RNGoogleMobileAdsCommon sendAdEvent:GOOGLE_MOBILE_ADS_EVENT_APP_OPEN
                                             requestId:requestId
                                                  type:GOOGLE_MOBILE_ADS_EVENT_LOADED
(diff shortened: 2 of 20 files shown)


ios/RNGoogleMobileAds/RNGoogleMobileAdsAppOpenModule.h
```diff
@@ -23,7 +23,6 @@
 
 @interface RNGoogleMobileAdsAppOpenModule : NSObject <RCTBridgeModule>
 
-@property(strong, nonatomic) GADAppOpenAd* appOpenAd;
-@property(strong, nonatomic) RNGoogleMobileAdsFullScreenContentDelegate* appOpenDelegate;
++ (void)removeAppOpen;
```
mikehardy, line 26: Is that correct? Is the only thing left in the header really the `removeAppOpen` symbol? If so, there is not an unneeded import I think of RNGoogleMobileAdsFullScreenContentDelegate, if not, does this work in local testing?

ios/RNGoogleMobileAds/RNGoogleMobileAdsAppOpenModule.m
```diff
@@ -21,6 +21,9 @@
 #import "RNGoogleMobileAdsCommon.h"
 #import "common/RNSharedUtils.h"
 
+static __strong GADAppOpenAd *appOpen;
+static __strong RNGoogleMobileAdsFullScreenContentDelegate *appOpenDelegate;
```
mikehardy, line 25: Ah, but I see they moved here. I may just not have the Objective-C skills to really review this (yet) and need to research more

ios/RNGoogleMobileAds/RNGoogleMobileAdsInterstitialModule.m
```diff
@@ -134,4 +133,15 @@ - (void)invalidate {
...
+  [RNGoogleMobileAdsInterstitialModule removeInterstitial:requestId];
+}
+
+#pragma mark -
+
++ (void)removeInterstitial:(nonnull NSNumber *)requestId {
+  [interstitialMap removeObjectForKey:requestId];
+  [interstitialDelegateMap removeObjectForKey:requestId];
+}
+
```
mikehardy, line 146: This is a) not this PRs fault and b) not something to take on in this PR but seeing the same code repeated over and over in both the java and the .m files makes me think "we have a lot of the same code for each ad type" which of course makes me think "we should probably have a base class that handles most of the "accounting" for keeping track of request ids and ads and destroying them etc, just with statics in the subclasses for the specific ad types and overrides / decorators when special behavior is needed

ios/RNGoogleMobileAds/RNGoogleMobileAdsAppOpenModule.m
wjaykim: Yes, I changed them to static var because other full screen ad modules declare them as static vars as well.

wjaykim: Yes, I changed them to static var because other full screen ad modules declare them as static vars as well.

src/hooks/useAppOpenAd.ts
```diff
@@ -37,7 +37,8 @@ export function useAppOpenAd(
   const [appOpenAd, setAppOpenAd] = useState<AppOpenAd | null>(null);
 
   useDeepCompareEffect(() => {
-    setAppOpenAd(() => {
+    setAppOpenAd(prevAd => {
```
mikehardy, line 40: if we null-check the type and it's not required, shouldn't the type of the setAppOpenAd function parameter be nullable? 🤔 

```suggestion
    setAppOpenAd(prevAd? => {
```

ios/RNGoogleMobileAds/RNGoogleMobileAdsInterstitialModule.m
wjaykim: Yes, in my previous library react-native-admob/admob, I removed duplicate codes via implementing common full screen ad class with generics, so we will have to refactor in the same manner for cleaner code.

src/ads/MobileAd.ts
```diff
@@ -208,12 +220,22 @@ export abstract class MobileAd implements MobileAdInterface {
...
     this._adEventsListeners.clear();
     this._adEventListenersMap.forEach((_, type, map) => {
       map.set(type, new Map());
     });
   }
 
+  public destroy() {
+    this._checkDestroyed();
+    const destroy = this._googleMobileAds.native[`${this._camelCaseType}Destroy`];
+    destroy(this._requestId);
```
mikehardy, line 233: This seems like it would work but it was confusing to the reader (me) to have the method called destroy, and the returned function also called destroy in local scope. At first I thought "Is this recursive?" 

suggest renaming to nativeDestroyCallback or something for easy reading comprehension

wjaykim: It makes sense.

src/hooks/useAppOpenAd.ts
wjaykim: The parameter becomes automatically nullable because type of `appOpenAd` state is `AppOpenAd | null`?

mikehardy (commented):
seems like a good idea - one of the things I like to do when I analyze performance optimizations is compare before and after

I looked for any existing react-native modules that spit out memory stats - including crucially native heap usage on android - but was unable to find any. It would be interesting at least in the logs  System.out something in debug modes before / after destroy - something like output of https://developer.android.com/reference/android/app/ActivityManager#getProcessMemoryInfo(int[]) pretty-printed 

Also, perhaps all the request ids should be `destroyed` with hook `onCatalystInstanceDestroy` like https://github.com/invertase/react-native-firebase/blob/f26482d6febc597f0a2ef3377a4c22821e33cdf0/packages/app/android/src/reactnative/java/io/invertase/firebase/common/ReactNativeFirebaseModule.java#L84-L88 to clean up on restarts of react-native - also I don't exactly remember the structure but if we override Activity anywhere should probably destroy with override in onStop of Activity ?

I am very strong on Android and very weak on iOS so I'm not sure if there are analogs on iOS to those methods...

What do you think?

mikehardy:
I should mention: one of the reasons I always try to get a look at before/after is because I am always surprised when it happens but frequently my performance optimization has zero effect 😆 - sometimes because of some small bug which when fixed then unlocks the optimization, and sometimes because what I thought was possible or faster was not possible or was not faster.

Without metrics, the impact on performance is unknown but the impact on future maintenance is known larger...something to avoid

wjaykim:
I'll investigate perf diff when available🙃

PR #130, opened 2022-05-03
fix(android, fluid): Initial attempt at supporting fluid ads

### LINKED ISSUE #120: Extend AdSize support to handle size=FLUID
Extend AdSize support to handle size=FLUID

_Originally posted by @mikehardy in https://github.com/invertase/react-native-google-mobile-ads/pull/105#discussion_r858917392_

djMax:
### Description

Fluid ads are not currently working in our React Native project, and it also appears are not supported in this module. I'm hoping that fixing it here will help us understand how to fix it in our project as well. This is an initial version, it's not even yet clear to me if it works, why it used to fail, etc. But I figured I'd start with a PR to get the ridicule out of the way early.

### Related issues

Fixes #120 

### Release Summary

Fluid ads report zero height in some cases, we now attempt to reconstruct the size properly.

### Checklist

  - [X] Yes
  - [X] `Android`
  - [X] `iOS` (only in that it is not relevant)
- My change includes tests;
  - [X] `e2e` tests added or updated in `__tests__e2e__`
  - [X] `jest` tests added or updated in `__tests__`
- This is a breaking change;
  - [X] No

### Test Plan

A test has been added, but I'm not yet convinced it's passing because I don't have control over the placements in the tests.

:fire:

### DIFF
--- a/RNGoogleMobileAdsExample/App.tsx
+++ b/RNGoogleMobileAdsExample/App.tsx
@@ -180,6 +180,41 @@ class BannerTest implements Test {
   }
 }
 
+class FluidTest implements Test {
+  getPath(): string {
+    return 'Fluid';
+  }
+
+  getTestType(): TestType {
+    return TestType.Interactive;
+  }
+
+  render(onMount: (component: any) => void): React.ReactNode {
+    return (
+      <View ref={onMount}>
+        <BannerAd
+          unitId={TestIds.FLUID}
+          size={BannerAdSize.FLUID}
+          requestOptions={{
+            requestNonPersonalizedAdsOnly: true,
+          }}
+        />
+      </View>
+    );
+  }
+
+  execute(component: any, complete: (result: TestResult) => void): void {
+    let results = new TestResult();
+    try {
+      // You can do anything here, it will execute on-device + in-app. Results are aggregated + visible in-app.
+    } catch (error) {
+      results.errors.push('Received unexpected error...');
+    } finally {
+      complete(results);
+    }
+  }
+}
+
 const rewarded = RewardedAd.createForAdRequest(TestIds.REWARDED, {
   requestNonPersonalizedAdsOnly: true,
   keywords: ['fashion', 'clothing'],
@@ -778,6 +813,7 @@ class GAMInterstitialTest implements Test {
 
 // All tests must be registered - a future feature will allow auto-bundling of tests via configured path or regex
 TestRegistry.registerTest(new BannerTest());
+TestRegistry.registerTest(new FluidTest());
 TestRegistry.registerTest(new AppOpenTest());
 TestRegistry.registerTest(new InterstitialTest());
 TestRegistry.registerTest(new RewardedTest());
--- a/src/TestIds.ts
+++ b/src/TestIds.ts
@@ -20,6 +20,7 @@ import { Platform } from 'react-native';
 export const TestIds = {
   APP_OPEN: '',
   BANNER: '',
+  FLUID: '',
   INTERSTITIAL: '',
   REWARDED: '',
   REWARDED_INTERSTITIAL: '',
@@ -33,6 +34,7 @@ export const TestIds = {
     android: {
       APP_OPEN: 'ca-app-pub-3940256099942544/3419835294',
       BANNER: 'ca-app-pub-3940256099942544/6300978111',
+      FLUID: 'ca-app-pub-3940256099942544/6300978111',
       INTERSTITIAL: 'ca-app-pub-3940256099942544/1033173712',
       REWARDED: 'ca-app-pub-3940256099942544/5224354917',
       REWARDED_INTERSTITIAL: 'ca-app-pub-3940256099942544/5354046379',
@@ -40,6 +42,7 @@ export const TestIds = {
     ios: {
       APP_OPEN: 'ca-app-pub-3940256099942544/5662855259',
       BANNER: 'ca-app-pub-3940256099942544/2934735716',
+      FLUID: 'ca-app-pub-3940256099942544/2934735716',
       INTERSTITIAL: 'ca-app-pub-3940256099942544/4411468910',
       REWARDED: 'ca-app-pub-3940256099942544/1712485313',
       REWARDED_INTERSTITIAL: 'ca-app-pub-3940256099942544/6978759866',
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsBannerAdViewManager.java
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsBannerAdViewManager.java
@@ -153,8 +153,12 @@ public class ReactNativeGoogleMobileAdsBannerAdViewManager
               // TODO size=FLUID is still not working
               left = 0;
               top = 0;
-              width = reactViewGroup.getWidth();
-              height = reactViewGroup.getHeight();
+              width = Math.max(
+                reactViewGroup.getWidth(),
+                adSize.getWidthInPixels(reactViewGroup.getContext()));
+              height = Math.max(
+                reactViewGroup.getHeight(),
+                adSize.getHeightInPixels(reactViewGroup.getContext()));
             } else {
               left = adView.getLeft();
               top = adView.getTop();
@@ -230,6 +234,9 @@ public class ReactNativeGoogleMobileAdsBannerAdViewManager
         ((AdManagerAdView) adView).setManualImpressionsEnabled(true);
       }
     } else {
+      if (sizes.contains(AdSize.FLUID) && sizes.size() == 1) {
+        isFluid = true;
+      }
       adView.setAdSize(sizes.get(0));
     }
 
--- a/type-test.ts
+++ b/type-test.ts
@@ -51,6 +51,7 @@ console.log(googleMobileAds.BannerAdSize.FLUID);
 console.log(googleMobileAds.BannerAdSize.FULL_BANNER);
 
 console.log(googleMobileAds.TestIds.BANNER);
+console.log(googleMobileAds.TestIds.FLUID);
 console.log(googleMobileAds.TestIds.INTERSTITIAL);
 console.log(googleMobileAds.TestIds.REWARDED);
 console.log(googleMobileAds.TestIds.BANNER);


mikehardy (commented):
Hmm 🤔  well first - I will ridicule someone taking the time post a PR exactly zero times in my life! Thanks for opening one up.

Unfortunately I'm not the best person to talk to about Ads domain knowledge, I'm more focused on keeping the repo running, so nothing I say here is for certain - *but* - the only documentation I see on fluid size pertains to "native" format ads, which do not actually support at all, yet

Here are the only doc refs I could dig up:
https://developers.google.com/ad-manager/mobile-ads-sdk/ios/native/native-styles#fluid_size
https://developers.google.com/ad-manager/mobile-ads-sdk/android/native/styles#fluid_size

Both of those are for the Google Ad Manager doc sets, nothing turned up in the non-Ad-Manager / regular-mobile-ads-sdk doc sets

We have opened communication with the only library we are aware of that supports native ads (https://ammarahm-ed.github.io/react-native-admob-native-ads/docs/introduction / https://github.com/ammarahm-ed/react-native-admob-native-ads/issues/274) but we have not begun the integration work yet and - crucially - it does not appear from the docs there that fluid size is supported yet? I wasn't able to see in a skim of the code either https://github.com/ammarahm-ed/react-native-admob-native-ads/tree/master/android/src/main/java/com/ammarahmed/rnadmob/nativeads

So, I could be totally wrong (maybe @wjaykim or @dylancom knows better?) but I *think* to support fluid sizes we need to support native ads first - likely coming soon - and *then* we can support the special fluid request type + associated display. I would love to be wrong in any of that because it means I'll learn something and I still have a lot to learn in the area at least...

wjaykim:
@mikehardy Actually, the name of Fluid size ad is native "style" ad. The style of the ad is created in web console of Ad Manager, and served in HTML code. The name is native, but is implemented with banner ad. 
The native ad which needs additional code currently is native "advanced" ad. 
So Fluid size ad can be implemented without implementing "native advanced" type.

src/TestIds.ts
```diff
@@ -33,13 +34,15 @@ export const TestIds = {
     android: {
       APP_OPEN: 'ca-app-pub-3940256099942544/3419835294',
       BANNER: 'ca-app-pub-3940256099942544/6300978111',
+      FLUID: 'ca-app-pub-3940256099942544/6300978111',
```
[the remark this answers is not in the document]
djMax: Yeah, I'm a little unclear too. But I'm pretty sure there is some sort of bug in the underlying google SDKs, and I'm just trying to prove it before sending it to them (again). getAdSize is returning what I think is the wrong value for these ads, for whatever reason.

djMax:
@wjaykim any idea what ad id I should be using to get a fluid ad back?

djMax:
Ok, I've used one of our "real" ad ids to get the fluid ad returned. And sure enough, the getAdSize() method returns the wrong thing - it returns 320x50_mb. The image that comes back is "tall" - if I manually set the height to the screen height, I see the whole ad. But adSize *InPixels returns 880x137 (definitely the wrong height). reactViewGroup getHeight returns 0, so also not useful.

djMax:
The google team pointed us at the Flutter SDK and said "go try that," but that didn't work either:

https://github.com/googleads/googleads-mobile-flutter/blob/main/packages/google_mobile_ads/android/src/main/java/io/flutter/plugins/googlemobileads/FluidAdManagerBannerAd.java#L54-L79

djMax:
I also see this discussion: https://github.com/invertase/react-native-firebase/issues/2590#issuecomment-547965751 Where is the relevant version of "requestLayout" in this infra?

djMax:
I tried subclassing ReactViewGroup but no luck. The core issue here is that I believe Google expects that the AdManagerAdView is going to resize itself when the content loads. I'm not sure what WE could be doing to stop that, but I do know that it doesn't happen. It requires us to set the width/height manually. But then we can't know what the height should be because we don't get accurate ad dimensions.

dylancom:
I'm not familiar with the "fluid" size, but isn't the adaptive banner format supposed to replace this? Which is the modern advised banner format...

wjaykim:
@dylancom Adaptive banner is to replace smart size ad. Fluid is different one.

dylancom:
@wjaykim I mean the "inline" adaptive banner which we don't support yet. https://developers.google.com/admob/ios/banner/inline-adaptive

The past AdMob docs used to advice about "fluid" banners, but that seems gone to me?

mikehardy:
So, as mentioned I am still coming up to speed on the whole "google mobile ads" area of domain knowledge, and I've been reading up on this. I find the support forum interesting to gather trends etc.

Things I learned 

- inline adaptive banners were released early access in 2020 so it makes sense that we don't support them here (yet!) as they were during the downtime on the library as we transitioned out of react-native-firebase/admob - and it does appear that they are the future so perhaps that is the real track to pursue? https://developers.google.com/admob/ios/banner/inline-adaptive

- fluid banner can have creative-specific problems that return height 0 ! this may be your issue @djMax  ? https://groups.google.com/g/google-admob-ads-sdk/c/a_vyBzKZRyg/m/jA6MOebPBwAJ - Some possible workarounds if we really want fluid to work on modern SDK verisons https://groups.google.com/g/google-admob-ads-sdk/c/0EJeAivB7Pk/m/29CN5GOyAQAJ 

- This is iOS-specific but if someone wants to check how we're loading fluid banners (if we support them?) on iOS that apparently changed in 2020 https://groups.google.com/g/google-admob-ads-sdk/c/kdNspRLdukU/m/1JI0RL3EAQAJ - and I have not investigated android yet but it may have an analogous change

This is more informative then actionable, as a comment - but to perhaps make it actionable: @djMax - would inline adaptive banners handle your use case? They appear like the future + have documentation whereas fluid banners seem to have just disappeared from the face of the SDK...

djMax:
We are going to look into it - as mentioned fluid ads DO work on iOS, but I'm not sure if adaptive ones do.

djMax:
Well looky here: https://github.com/Freddixx/react-native-ad-manager/pull/7/files

mikehardy:
That's a great find - was not aware of that library at all - the ad module space is so heavily fragmented it's pretty crazy!
I'll ask them for confirmation that on a banner ad on android, in their experience, *any* event means it is a fluid event. That seems a little open-ended to me but I imagine they have experience...

djMax:
I'm trying to recreate their fix but not currently getting a call to appevent. We do already subscribe to that event.

mikehardy:
Hopefully they have the answers and/or want to join forces here, always great to have more maintainers as I think it makes for a higher quality module and healthier maintainer experience...

djMax:
I found a Google sample [googleads/googleads-mobile-android-examples](https://github.com/googleads/googleads-mobile-android-examples/blob/master/java/advanced/APIDemo/app/src/main/java/com/google/android/gms/example/apidemo/AdManagerFluidSizeFragment.java#L61) that works. Their structure is very different so I'm not really sure what to take from it other than knowing the problem is not strictly in the underlying sdk.

djMax:
Interesting discovery. The forums asked me to verify that the parent view had a width, because I guess that's required. When we call loadAd, the ReactViewGroup has no parent, and thus cannot have a width. By the time "adLoaded" comes around, we have a parent, but I'm thinking that might be too late. I don't know a ton about this dance, but when does the view group get a parent?

djMax:
https://stackoverflow.com/questions/39836356/react-native-resize-custom-ui-component

I think that React Native does something that causes traditional LayoutParams based layouts to fail. I'm really not sure what to do to fix this, without getting Google to fix their mess.

mikehardy:
I think you mean facebook (or the community? I think majority of PRs are from the community) as google would be fixing flutter ;-). Quite the digging there and it does look like you're on to something, I've read the issues linked from that stackoverflow - it fits the evidence that the android sample works while it doesn't work in RN either.

All the code in RN for layout is open source too - may be possible to pinpoint it and either fix it or workaround it or something in this module still...it appears two modules already did?

The extracted modules linked there are still doing it:

https://github.com/react-native-picker/picker/blob/fcd0bbf4ff11cfcf2e2ae9dc519d46292c5c24a7/android/src/main/java/com/reactnativecommunity/picker/ReactPicker.java#L123-L142

That said - someone else just adapted that fix for fluid ads as well - https://github.com/zjkuang/OpenWrapInRN/commit/0793e822952c53f23522358106d1e2c8b9563761 / https://github.com/zjkuang/OpenWrapInRN/commit/0793e822952c53f23522358106d1e2c8b9563761#diff-fd32a82600be722549cdc575086235f4479ec80a30449d3b96d89ec474018896R282-R299

djMax:
I meant Google. The mess is that the ad size is unavailable on Android, but readily accessible on iOS. A very simple fix for them (and one that makes them money) vs a very complex one for React Native

mikehardy:
fair enough! maybe we can help fix it if they open source the admob SDK (as if... ;-) )

djMax:
Yeah, that would've made this almost trivial. That, or Google actually supporting the library. Anyhow, we're going to see if this module (https://github.com/Freddixx/react-native-ad-manager) works for our setup, and if so try to figure out what we missed when trying to adopt their approach.

mikehardy:
Great - sorry I/we haven't been able to help you just snipe it here so far! You've been incredibly resourceful looking for the missing magic code so far. Obviously we're ready to help if you can get something tested+working. Good luck

djMax:
Thanks, I appreciate the help so far. Very frustrating when you just want to step into some code but the vendor intentionally obfuscates it, combined with lack of support. But we press on!

djMax:
Sadly, the other library seems to have the same problem.

RNGoogleMobileAdsExample/App.tsx
```diff
@@ -180,6 +180,41 @@ class BannerTest implements Test {
...
+  }
+
+  getTestType(): TestType {
+    return TestType.Interactive;
+  }
+
+  render(onMount: (component: any) => void): React.ReactNode {
+    return (
+      <View ref={onMount}>
+        <BannerAd
```
wjaykim, line 195:
```suggestion
        <GAMBannerAd
```

RNGoogleMobileAdsExample/App.tsx
```diff
@@ -180,6 +180,41 @@ class BannerTest implements Test {
...
+          unitId={TestIds.FLUID}
+          size={BannerAdSize.FLUID}
```
wjaykim, line 197:
```suggestion
          sizes={[BannerAdSize.GAM_NATIVE]}
```

src/TestIds.ts
```diff
@@ -20,6 +20,7 @@ import { Platform } from 'react-native';
 export const TestIds = {
   APP_OPEN: '',
   BANNER: '',
+  FLUID: '',
```
wjaykim, line 23:
```suggestion
  GAM_NATIVE: '/6499/example/native',
```

src/TestIds.ts
wjaykim, line 37:
```suggestion
```

src/TestIds.ts
```diff
@@ -33,13 +34,15 @@ export const TestIds = {
...
       BANNER: 'ca-app-pub-3940256099942544/6300978111',
+      FLUID: 'ca-app-pub-3940256099942544/6300978111',
       INTERSTITIAL: 'ca-app-pub-3940256099942544/1033173712',
       REWARDED: 'ca-app-pub-3940256099942544/5224354917',
       REWARDED_INTERSTITIAL: 'ca-app-pub-3940256099942544/5354046379',
     },
     ios: {
       APP_OPEN: 'ca-app-pub-3940256099942544/5662855259',
       BANNER: 'ca-app-pub-3940256099942544/2934735716',
+      FLUID: 'ca-app-pub-3940256099942544/2934735716',
```
wjaykim, line 45:
```suggestion
```

wjaykim (commented):
Maybe we want to test FLUID with GAMBanner and GAM_NATIVE test ad id.

djMax:
Without a doubt, a big part of the problem is adtech is super complicated. Do I have it right that "fluid" just means "the ad will have a set size, but you don't specify it in the request - the ad server decides which one to give you"? And I notice that these ad requests all seem to return HTML, even if it's just an image ad. That HTML is rendered in a WebView in some way, and then determines its desired size, somehow. Is that all right?

PR #132, opened 2022-05-04
feat: add inline adaptive banners support

### LINKED ISSUE #128: Support inline adaptive banners
Our current adaptive banner implementation seems to focus around "anchored adaptive banners".
There is also an "inline adaptive banner" variant we should support.

https://developers.google.com/admob/ios/banner/anchored-adaptive
https://developers.google.com/admob/ios/banner/inline-adaptive

dylancom:
### Description
Fixes #128

### DIFF
--- a/RNGoogleMobileAdsExample/App.tsx
+++ b/RNGoogleMobileAdsExample/App.tsx
@@ -146,8 +146,14 @@ class InterstitialTest implements Test {
 }
 
 class BannerTest implements Test {
+  bannerAdSize: BannerAdSize | string;
+
+  constructor(bannerAdSize) {
+    this.bannerAdSize = bannerAdSize;
+  }
+
   getPath(): string {
-    return 'Banner';
+    return this.bannerAdSize;
   }
 
   getTestType(): TestType {
@@ -159,7 +165,7 @@ class BannerTest implements Test {
       <View ref={onMount}>
         <BannerAd
           unitId={TestIds.BANNER}
-          size={BannerAdSize.ADAPTIVE_BANNER}
+          size={this.bannerAdSize}
           requestOptions={{
             requestNonPersonalizedAdsOnly: true,
           }}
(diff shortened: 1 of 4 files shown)


src/BannerAdSize.ts
```diff
@@ -43,10 +43,21 @@ export enum BannerAdSize {
...
   /**
-   * A (next generation) dynamically sized banner that is full-width and auto-height.
+   * @deprecated Use `ANCHORED_ADAPTIVE_BANNER` instead.
    */
   ADAPTIVE_BANNER = 'ADAPTIVE_BANNER',
 
+  /**
+   * A (next generation) dynamically sized banner that is full-width and auto-height.
+   */
+  ANCHORED_ADAPTIVE_BANNER = 'ANCHORED_ADAPTIVE_BANNER',
```
mikehardy, line 53: Something occurred to me when I saw all these sizes, in our example app we are only every using ADAPTIVE_BANNER - https://github.com/invertase/react-native-google-mobile-ads/blob/eb149a62f26968c81646edd865567a9ee42936d6/RNGoogleMobileAdsExample/App.tsx#L162 - but we've code that branches on ad size, so it seems clear we should iterate on banner sizes as well in the App.tsx for example

Do you think it would be reasonable to have an array of all  banner sizes to show, then foreach banner size create + register one test component so each are visible? 

I wonder if that would have caught the fluid banner size issue on android actually! May be useful to actually draw a box with borderColor !== default and borderWidth >= 1 so it's clear whether it is showing or not, and my hypothesis is that an example iteration over all types would show the fluid issue, while demonstrating the three adaptive banner types work

What do you think?

mikehardy (commented):
This looks really clean just a question on example demonstration

src/BannerAdSize.ts
dylancom: I think that's a good idea and added a test for every banner.

RNGoogleMobileAdsExample/App.tsx
```diff
@@ -777,7 +783,9 @@ class GAMInterstitialTest implements Test {
 }
 
 // All tests must be registered - a future feature will allow auto-bundling of tests via configured path or regex
-TestRegistry.registerTest(new BannerTest());
+Object.keys(BannerAdSize).forEach(bannerAdSize => {
+  TestRegistry.registerTest(new BannerTest(bannerAdSize));
+});
```
mikehardy, line 788: Slick! won't even need updates as new sizes added, love it

mikehardy (approved):
To me this looks like a really nice change - tiny + clean code change and the example showing all banners. +1 / merge + release at will as far as I'm concerned, when you're happy with it. Nice!

Issue #133, opened 2022-05-06
🔥 [🐛] The provided view controller is already presenting another view controller


Bartozzz:
## Issue

Related to https://github.com/invertase/react-native-google-mobile-ads/issues/47. 

**Context:**
I am a hobbyist game developer making games w/ Expo. [`expo-ads-admob`](https://docs.expo.dev/versions/v45.0.0/sdk/admob/) will get deprecated in SDK 46 and the official documentation encourages to use `react-native-google-mobile-ads` instead:

> Deprecated. This module will be removed in SDK 46. There will be no replacement that works with the classic build service (expo build) because [the classic build service has been superseded by EAS Build](https://blog.expo.dev/turtle-goes-out-to-sea-d334db2a6b60). With EAS Build and [Development Builds](https://docs.expo.dev/development/introduction/), you should use [react-native-google-mobile-ads](https://github.com/invertase/react-native-google-mobile-ads) instead.

**Problem:**
Interstitial and Rewarded Ads cannot be shown inside modals (including `react-navigation` stacks rendered with `screenOptions={{ presentation: "modal" }}`). Trying to show an advertisement inside a modal results in the following error:

```
The provided view controller is already presenting another view controller
```

Are there plans to fix this? Are there any workarounds that do not require big refactors? Note that it was working without any issues in `expo-ads-admob`. Displaying ads on-top of modals is a popular practice in game development, for example when showing a _"You won"_ modal at the end of a level.

---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
{
  "scripts": {
    "start": "expo start --dev-client",
    "android": "expo run:android",
    "ios": "expo run:ios"
  },
  "dependencies": {
    "@react-navigation/elements": "^1.3.3",
    "@react-navigation/native": "^6.0.2",
    "@react-navigation/native-stack": "^6.6.2",
    "expo": "~44.0.0",
    "expo-asset": "~8.4.4",
    "expo-constants": "~13.0.0",
    "expo-device": "~4.1.0",
    "react": "17.0.1",
    "react-dom": "17.0.1",
    "react-native": "0.64.3",
    "react-native-gesture-handler": "~2.1.0",
    "react-native-google-mobile-ads": "^6.2.0",
    "react-native-reanimated": "~2.3.1",
    "react-native-safe-area-context": "3.3.2",
    "react-native-screens": "~3.10.1",
    "react-native-web": "0.17.1"
  },
[...]

PR #134, opened 2022-05-06
fix(ios): already presenting another view controller

dylancom:
### Description

Should fix: #133
Needs thorough testing.
Code borrowed from: https://github.com/expo/expo/blob/d84b7ae9c09e1377f415004167c70ea51b888d56/packages/expo-modules-core/ios/EXUtilities.m#L40


mikehardy (approved):
Fantastic! Looks like a clean way to get the topmost controller, and has verified success from original reporter. Can't beat that

Issue #137, opened 2022-05-09
🔥[🐛] Misleading TypeScript types


Bartozzz:
e## Issue

In [`RequestConfiguration` interface](https://github.com/invertase/react-native-google-mobile-ads/blob/469b3be4197545c0c6f1da248539d94b864d9be9/src/types/RequestConfiguration.ts) every property is optional, so in theory the following piece of code is considered correct:

```tsx
await mobileAds().setRequestConfiguration({
  // Update all future requests suitable for parental guidance:
  maxAdContentRating: MaxAdContentRating.PG,

  // An array of test device IDs to allow:
  testDeviceIdentifiers: isDev ? ["EMULATOR"] : undefined,
});
```

In practice, it will throw an error on the production release:

```
googleMobileAds.setRequestConfiguration(*) 'requestConfiguration.testDeviceIdentifiers' expected an array value
```

This is because this library performs [`hasOwnProperty` check](https://github.com/invertase/react-native-google-mobile-ads/blob/469b3be4197545c0c6f1da248539d94b864d9be9/src/validateAdRequestConfiguration.ts#L62=) which returns `true` if a given property exists even if its value is `undefined`. This can lead to unexpected, sometimes hard to spot bugs. In my case, some assets were not loading properly because of the piece of code above (as you can see, `testDeviceIdentifiers` value is invalid only for production release).

I'd recommend either updating the TypeScript types or update the check with something like:

```tsx
import { isUndefined, isNull } from "./validate";

function isPropertySet(target: unknown, property: PropertyKey) {
  return hasOwnProperty(target, property) && !isUndefined(target[property]) && !isNull(target[property]);
}
```

LMK If this solutions suites you, I can open a PR.

## Environment

<details><summary>Click To Expand</summary>
<p>

- **Platform that you're experiencing the issue on**:
  - [ ] iOS
  - [ ] Android
  - [ ] **iOS** but have not tested behavior on Android
  - [ ] **Android** but have not tested behavior on iOS
  - [x] Both
- **Are you using `TypeScript`?**
  - `~4.3.5`

</p>
</details>

mikehardy:
Interesting! I'd like for this library to still be useful even for those that don't use typescript, so it seems the javascript check should be updated.

That said, I'm a typescript person so it should "feel" correct in typescript usage - in that case since your proposed javascript update check would allow the undefined (without crashing) it seems the optional spec for that property would then be working as expected.

If that analysis matches your thinking - a PR would be great!

PR #138, opened 2022-05-09
fix: allow undefined/null configuration properties

Bartozzz:
### Description

In [`RequestConfiguration` interface](https://github.com/invertase/react-native-google-mobile-ads/blob/469b3be4197545c0c6f1da248539d94b864d9be9/src/types/RequestConfiguration.ts) every property is optional, so in theory the following piece of code is considered correct:

```tsx
await mobileAds().setRequestConfiguration({
  testDeviceIdentifiers: undefined,
});
```

In practice, it will throw an error on the production release:

```
googleMobileAds.setRequestConfiguration(*) 'requestConfiguration.testDeviceIdentifiers' expected an array value
```

This is because this library performs [`hasOwnProperty` check](https://github.com/invertase/react-native-google-mobile-ads/blob/469b3be4197545c0c6f1da248539d94b864d9be9/src/validateAdRequestConfiguration.ts#L62=) which returns `true` if a given property exists event if its value is `undefined`.

I created a `isPropertySet` to check not only if the property is set, but also if it's not `null` or `undefined`. With this change the example code above will not throw, because `testDeviceIdentifiers` will get ignored.

### Related issues

Fixes https://github.com/invertase/react-native-google-mobile-ads/issues/137

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
  - [x] `e2e` tests added or updated in `__tests__e2e__`
- This is a breaking change;
  - [ ] Yes
  - [x] No
  
  🔥

### DIFF
--- a/src/common/index.ts
+++ b/src/common/index.ts
@@ -17,7 +17,7 @@
 
 import { Platform } from 'react-native';
 import * as Base64 from './Base64';
-import { isString } from './validate';
+import { isString, isNull, isUndefined } from './validate';
 
 export * from './id';
 export * from './path';
@@ -67,6 +67,12 @@ export function isError(value: unknown) {
 
 export function hasOwnProperty(target: unknown, property: PropertyKey) {
   return Object.hasOwnProperty.call(target, property);
+}
+
+export function isPropertySet(target: any, property: PropertyKey) {
+  return (
+    hasOwnProperty(target, property) && !isUndefined(target[property]) && !isNull(target[property])
+  );
 }
 
 /**
--- a/e2e/admob.e2e.js
+++ b/e2e/admob.e2e.js
@@ -49,6 +49,11 @@ describe('googleAds', function () {
           maxAdContentRating: googleAds.MaxAdContentRating.G,
         });
       });
+
+      it('accepts undefined or null', async function () {
+        await googleAds.setRequestConfiguration({ maxAdContentRating: undefined });
+        await googleAds.setRequestConfiguration({ maxAdContentRating: null });
+      });
     });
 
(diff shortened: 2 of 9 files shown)


src/common/index.ts
```diff
@@ -69,6 +69,14 @@ export function hasOwnProperty(target: unknown, property: PropertyKey) {
   return Object.hasOwnProperty.call(target, property);
 }
 
+export function isPropertySet(target: unknown, property: PropertyKey) {
+  return (
+    hasOwnProperty(target, property) &&
+    !isUndefined((target as Record<PropertyKey, unknown>)[property]) &&
+    !isNull((target as Record<PropertyKey, unknown>)[property])
```
wjaykim, line 76: Is checking `isNull` correct? If the value is `null`, it means the value is set to `null`, doesn't it?

wjaykim (changes requested):
Property can be set to `null`, it doesn't mean the property is not set, so maybe we shouldn't check if the property is `null` in `isPropertySet`? What do you think?

src/common/index.ts
mikehardy: I agree with this one - I liked the discussion version we had on linked issue where it was looking for undefined and letting that go, but null is a specific (anti-)value.

But more than anything, thanks for noticing this and taking the time to post a PR, really appreciated.

mikehardy (approved):
This looks good to me, thanks @Bartozzz - @wjaykim was involved in review too so requesting another look but I'm +1 for merge+launch personally, Cheers!

wjaykim:
Looks good to me, too. 🚀

PR #141, opened 2022-05-10
feat: Get adaptive banner height

Mikenso:
### Description
At current moment when we use Anchored Adaptive banner and get loading error BannerAd does not take up space on the screen, this may cause to accidental clicks when banner finally loads and google can restrict admob account .
To prevent users from accidental clicks we should reserve screen space for banner before banner shows up, but for anchored adaptive banner  height can variate from 50px to 15% of screen height so we need to get banner height for every device and apply to styles of banner container

### Checklist

  - [X] Yes
  - [X] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

### DIFF
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeAppModule.java
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeAppModule.java
@@ -64,7 +64,7 @@ public class ReactNativeAppModule extends ReactNativeModule {
     int dpWidth = (int) (displayMetrics.widthPixels / displayMetrics.density);
 
     AdSize adaptiveBannerSize =
-      AdSize.getCurrentOrientationAnchoredAdaptiveBannerAdSize(getContext(), dpWidth);
+        AdSize.getCurrentOrientationAnchoredAdaptiveBannerAdSize(getContext(), dpWidth);
     return adaptiveBannerSize.getHeight();
   }
 
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeAppModule.java
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeAppModule.java
@@ -17,10 +17,12 @@ package io.invertase.googlemobileads;
  *
  */
 
+import android.util.DisplayMetrics;
 import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactMethod;
 import com.facebook.react.bridge.ReadableMap;
+import com.google.android.gms.ads.AdSize;
 import io.invertase.googlemobileads.common.RCTConvert;
 import io.invertase.googlemobileads.common.ReactNativeEvent;
 import io.invertase.googlemobileads.common.ReactNativeEventEmitter;
@@ -53,6 +55,17 @@ public class ReactNativeAppModule extends ReactNativeModule {
     // RCTConvertFirebase.reactNativeAppToWritableMap(reactNativeApp);
     // promise.resolve(reactNativeAppMap);
     promise.resolve(options);
+  }
+
+  @ReactMethod(isBlockingSynchronousMethod = true)
+  public int getCurrentOrientationAnchoredAdaptiveBannerHeight() {
+    DisplayMetrics displayMetrics = getContext().getResources().getDisplayMetrics();
+
+    int dpWidth = (int) (displayMetrics.widthPixels / displayMetrics.density);
+
+    AdSize adaptiveBannerSize =
+      AdSize.getCurrentOrientationAnchoredAdaptiveBannerAdSize(getContext(), dpWidth);
+    return adaptiveBannerSize.getHeight();
   }
 
   @ReactMethod
--- a/src/ads/BannerAd.tsx
+++ b/src/ads/BannerAd.tsx
@@ -16,9 +16,15 @@
  */
 
 import React from 'react';
+import { NativeModules } from 'react-native';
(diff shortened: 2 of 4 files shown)


android/src/main/java/io/invertase/googlemobileads/ReactNativeAppModule.java
```diff
@@ -55,6 +57,17 @@ public void initializeApp(ReadableMap options, ReadableMap appConfig, Promise pr
     promise.resolve(options);
   }
 
+  @ReactMethod(isBlockingSynchronousMethod = true)
```
mikehardy, line 60: I have gotten a lot of negative feedback from module users historically when I've used this annotation - it has the side effect of making it so that the JSI debugging interface won't work, as you cannot use synchronous methods and the JSI. Is this strictly necessary, with no way around it? It would be the first instance of it in this module and going from 0 to 1 is a big step...

src/ads/BannerAd.tsx
```diff
@@ -16,9 +16,15 @@
...
 import { BannerAdProps } from '../types/BannerAdProps';
 import { BaseAd } from './BaseAd';
 
 export function BannerAd({ size, ...props }: BannerAdProps) {
   return <BaseAd sizes={[size]} {...props} />;
 }
+
+export function getCurrentOrientationAnchoredAdaptiveBannerHeight() {
+  const { RNAppModule } = NativeModules;
+  return RNAppModule.getCurrentOrientationAnchoredAdaptiveBannerHeight();
```
mikehardy, line 29: If you call into the native module without a platform check, and without an iOS implementation, this will cause a redbox error / crash on iOS. Is there a default implementation we should implement for non-android platforms? Or do we need an implementation for iOS as well?

mikehardy (changes requested):
Oh very interesting! This is a valid user concern so I can understand why google would have this stance:

> At current moment when we use Anchored Adaptive banner and get loading error BannerAd does not take up space on the screen, this may cause to accidental clicks when banner finally loads and google can restrict admob account .

And of course we don't want any excuses to restrict the account, it's hard enough complying with everything I think.

So this is great, thanks for posting it

I had a couple questions about implementation but it seems like a real need, and the implementation for the java side at least does seem to address the need

android/src/main/java/io/invertase/googlemobileads/ReactNativeAppModule.java
Mikenso: If we make asynchronous  - we will face another inconvenience - we will need somehow to get this value before showing banner or there still be fast content offset

src/ads/BannerAd.tsx
Mikenso: Yes, We should implement this for IOS too, but since I am not familiar with the objective-s and did not know whether the feature would be in demand or rejected, I saved my energy

android/src/main/java/io/invertase/googlemobileads/ReactNativeAppModule.java
mikehardy: Hmm - if it is just order of operations then I'd rather it be async with good documentation on how to call this first and/or make a best guess first then call this etc to block the space out, rather than forced sync and breaking JSI usage - all I can tell you from experience is that you really don't want to do that if you can avoid it. You (I? we? :laughing: ) will get a *lot* of complaints and invariably you'll have a sync and async version which is just a pain

src/ads/BannerAd.tsx
mikehardy: I'll wait to see what others think to see if they also like it then, so as not to waste any time if I'm the only fan :-)

dylancom:
Would like to see a reproduction in our example.
Just did a test and if I put in an incorrect ad unit, which will trigger an error, the banner still takes up the appropriate height.
Schermafbeelding 2022-05-11 om 08 42 59

I wrapped a div with backgroundColor 'black' around the banner to make it obvious.

android/src/main/java/io/invertase/googlemobileads/ReactNativeAppModule.java
dylancom: We also already have a similar static method: https://github.com/invertase/react-native-google-mobile-ads/blob/main/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsCommon.java#L43

src/ads/BannerAd.tsx
dylancom: We also already have a similar static method: https://github.com/invertase/react-native-google-mobile-ads/blob/main/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsCommon.java#L43

android/src/main/java/io/invertase/googlemobileads/ReactNativeAppModule.java
```diff
@@ -55,6 +57,17 @@ public void initializeApp(ReadableMap options, ReadableMap appConfig, Promise pr
...
+  public int getCurrentOrientationAnchoredAdaptiveBannerHeight() {
```
dylancom, line 61: We already have a similar static method: https://github.com/invertase/react-native-google-mobile-ads/blob/main/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsCommon.java#L43

wjaykim:
I think the issue is resolved with #146, so closing this PR.

mikehardy:
Excellent @wjaykim and thanks a bunch for working with on this @Mikenso ! Really appreciated

Issue #109, opened 2022-04-08
Investigate iOS rewarded ad no-fill errors in example app


mikehardy:
I asked the SDK forum about iOS rewarded ad no-fills: https://groups.google.com/g/google-admob-ads-sdk/c/5ixRSHOkhHE

_Originally posted by @mikehardy in https://github.com/invertase/react-native-google-mobile-ads/issues/108#issuecomment-1093381741_

mikehardy:
Okay, this was the response from the forum:

> Hi Mike,
> 
> Thank you for reaching out to us.
> 
> I've checked our test ad unit for rewarded ads in our Sample App iOS, and I was able to get ads on it. It means that this has something to do in your implementation. Since your implementation is react native, I would suggest reaching out to their support team as we only support implementation for Android, iOS, Unity and Flutter.
> 
> Regards,
> Google Logo 	
> Princess Pamela
> Mobile Ads SDK Team

Which is great! I mean, it means our iOS rewarded ad implementation is a problem but a) we could see it! our example app was sufficient to detect the issue, that's a good result and b) it's working in other integrations so not ours, meaning the problem is localized here.

*Something* is wrong in our rewarded ads implementation for iOS, and it doesn't appear to be the native SDK or the ad unit ID.

So we have a valid test rig to see failure, and we are in control of the issue in our code.

This just needs internal triage + dev effort now if anyone has time to pick it up

mikehardy:
@wjaykim you mentioned that trying to load two rewarded ads at once could cause problems (in #135) - I wonder if that could be the problem in the example app? What's strange is it works for android but not for iOS (the opposite of #135). This one seems strange since it did not reproduce for the support team on a pure native app, it works on android, and I swear I see an "advertisement amount of data traffic" (vague but trust me) download when I enter the rewarded ads hook implementation for instance, and it loads one. But on ad inspector it says no-fill :thinking:

wjaykim:
But in my local environment, rewarded ad in example app works both in Android and iOS 🤔

mikehardy:
Hahah - of *course* it does. Couldn't be a reproducible problem, that would be too easy ;-).
I'll examine it from an environment difference angle though, simply knowing it works for you is really helpful.
Any information you can give on what you used (say, `react-native info` or similar) might help

wjaykim:
Sure, here it is,
```
System:
    OS: macOS 12.3.1
    CPU: (8) arm64 Apple M1 Pro
[...]

PR #167, opened 2022-06-12
refactor: refactor full screen ads

wjaykim:
### Description

Refactor native full screen ads modules with Kotlin and Swift, and also generic expressions to minimize duplicate codes.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---


android/build.gradle
```diff
@@ -12,13 +12,15 @@ buildscript {
 
     dependencies {
       classpath("com.android.tools.build:gradle:7.0.4")
+      classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:${safeExtGet('kotlinVersion', '1.6.10')}"
```
mikehardy, line 15: Not the most important thing, but kotlin is at 1.7.0 now, any reason to pick 1.6.10 as the fallback specifically, or could we use 1.7.0 so we're at least current to start?

mikehardy (approved):
Seems like  a standard conversion more or less. Probably best to do a major version more as a "this should not break anything but just in case" release. I mentioned one tiny thing about kotlin version but in my current testing 1.7.0 of kotlin has been compatible with everything so should be fine?

Going to pull this locally and test it, if there's no regression then it's just resolution of the kotlin version and merge+release I think?

mikehardy:
Testing notes, none of these are regressions, but I thought they were worth noting:

1) I still see no-fill for rewarded / rewardedinterstitial on iOS despite network traffic indicating an ad came down #109 
2) iOS shows skyscraper but for some reason android does not

Everything else looks fine, and it all looks the same on main or this branch, so no reason to block it

android/build.gradle
mikehardy: will stick with this for now - I want to get this in prior to any work updating to new SDK #175

mikehardy:
No worries ! It's still open source right? Should not feel guilty when you're busy but it's always great to see you around. Cheers

PR #212, opened 2022-08-30
fix(android, sdk): passing custom targeting as array of string instead of string

AbdelhalimAhmed:
### Description

When passing customTargeting as Array of string instead of string the app is crashed.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

https://developers.google.com/android/reference/com/google/android/gms/ads/admanager/AdManagerAdRequest.Builder#public-admanageradrequest.builder-addcustomtargeting-string-key,-liststring-values
---

### DIFF
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsCommon.java
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsCommon.java
@@ -178,7 +178,7 @@ public class ReactNativeGoogleMobileAdsCommon {
 
       for (Map.Entry<String, Object> entry : customTargeting.entrySet()) {
         String key = entry.getKey();
-        String value = (String) entry.getValue();
+        ArrayList<String> value = (ArrayList<String>) entry.getValue();
         builder.addCustomTargeting(key, value);
       }
     }
--- a/src/types/RequestOptions.ts
+++ b/src/types/RequestOptions.ts
@@ -70,7 +70,7 @@ export interface RequestOptions {
    *
    * Takes an array of string key/value pairs.
    */
-  customTargeting?: { [key: string]: string };
+  customTargeting?: { [key: string]: string[] };
 
   /**
    * Sets the request agent string to identify the ad request's origin. Third party libraries that reference the Mobile
--- a/src/types/RequestOptions.ts
+++ b/src/types/RequestOptions.ts
@@ -70,7 +70,7 @@ export interface RequestOptions {
    *
    * Takes an array of string key/value pairs.
    */
-  customTargeting?: { [key: string]: string[] };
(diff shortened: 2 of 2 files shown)


wjaykim:
Hi! Thank you for contributing to our repo. 
Does this change work with iOS, too?

src/types/RequestOptions.ts
```diff
@@ -70,7 +70,7 @@ export interface RequestOptions {
    *
    * Takes an array of string key/value pairs.
    */
-  customTargeting?: { [key: string]: string };
+  customTargeting?: { [key: string]: string[] };
```
wjaykim, line 73: When we makes change to type, it is breaking change. To support previous versions, I think it should be union type of `string | string[]`.

wjaykim (changes requested):

AbdelhalimAhmed:
> 

yes, works fine

src/types/RequestOptions.ts
AbdelhalimAhmed: updated

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsCommon.java
```diff
@@ -178,7 +178,7 @@ public static AdManagerAdRequest buildAdRequest(ReadableMap adRequestOptions) {
 
       for (Map.Entry<String, Object> entry : customTargeting.entrySet()) {
         String key = entry.getKey();
-        String value = (String) entry.getValue();
+        ArrayList<String> value = (ArrayList<String>) entry.getValue();
```
mikehardy, line 181: I'm having a difficult time understanding how this does not just convert the crash from being:

"when passing array of strings it crashes" to

"when passing single string it crashes"

That is, this just makes it crash for the other case now. It appears both types of input should be accepted (`String` and `ArrayList<String>`) and you need to do a runtime check like `if (value instanceof String) ...` with conditional cast

mikehardy (changes requested):
Not sure this doesn't cause a different kind of crash, as commented near line of code

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsCommon.java
wjaykim: Yes, I think so, too.

mikehardy: Just a note @AbdelhalimAhmed that this looks close to going in, but this comment must be addressed as it appears it will cause a crash in the opposite direction of your fix, and both cases need to be handled...

PR #257, opened 2022-10-27
fix(ios): correct typo in ios_config.sh

Xanewok:
Tweak the ios_config.sh error to mention the `GADApplicationIdentifier` key from the Info.plist using the correct value - it may help with Ctrl-C troubleshooting.


mikehardy (approved):
Ha! :facepalm: good catch, thank you for posting a fix

PR #269, opened 2022-11-10
Use activity instead of context for init

TaylorDale:
This is required by some mediation partners (so far as I have found, partner: Ironsource).

Android only.
See error code 102: 
https://developers.google.com/admob/android/mediation/ironsource#step_4_additional_code_required

### DIFF
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsModule.java
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsModule.java
@@ -120,7 +120,7 @@ public class ReactNativeGoogleMobileAdsModule extends ReactNativeModule {
   @ReactMethod
   public void initialize(Promise promise) {
     MobileAds.initialize(
-        getApplicationContext(),
+        getCurrentActivity(),
         new OnInitializationCompleteListener() {
           @Override
           public void onInitializationComplete(InitializationStatus initializationStatus) {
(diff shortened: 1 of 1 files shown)


mikehardy (changes requested):
:thinking:  this is a bit of a difficult one.

react-native is semi-famous (among module maintainers) for the current Activity being unexpectedly null.

The application context always exists, but since React Native is async across the javascript bridge (as it stands now, not sure about New Architecture / Fabric) it is possible that the native code is still executing after some application lifecycle changes that mean the Activity has actually been destroyed along with the bridge, and getCurrentActivity returns null.

It's rare, but I have seen it multiple times myself over a few years.

So I understand the need for your change, and I am not against using getCurrentActivity, but you must use it in a null-safe way.

I propose setting a local variable at the top of the method to the result of getCurrentActivity, testing it for null, rejecting the Promise if it is null, then using the local variable in the rest of the method.

That way you get your Activity but we're null safe for the rare issue I describe.

What do you think?

TaylorDale:
Well.. can't say you didn't warn me. 😂

image

mikehardy:
Sometimes it's no fun to be correct :laughing:

mikehardy:
Friendly nudge on this one - I think there's a path forward here that gets it working for you but is also null safe

TaylorDale:
Sorry I've had patch-package operating for a bit on this one so I forgot to update it - cheers

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsModule.java
```diff
@@ -120,7 +120,8 @@ private RequestConfiguration buildRequestConfiguration(ReadableMap requestConfig
   @ReactMethod
   public void initialize(Promise promise) {
     MobileAds.initialize(
-        getApplicationContext(),
+        /* Use activity when available - preferrable for mediated networks*/
+        (getCurrentActivity() != null) ? getCurrentActivity() : getApplicationContext(),
```
DoctorJohn, line 124: Is the second call to getCurrentActivity guaranteed the return the same result as the first one?

PR #286, opened 2022-12-06
fix(ios): adaptive banner should take safe areas into account

dylancom:
### Description

Currently adaptive banner ads can overlap with the notch / dynamic island in landscape mode.
They should take safe areas into account.

Source:
https://developers.google.com/admob/ios/banner/anchored-adaptive#implementation_notes


mikehardy (approved):
LGTM :-), thanks @dylancom

PR #296, opened 2022-12-24
feat: inline adaptive banner with custom width

dylancom:
### Description

Currently an adaptive banner always takes the screen width (with taking into account safe areas).
There are multiple users that would like to use a custom width and seem to fall back to trying out "FLUID" sizing?
Currently you can define a custom size "300x200" to get a custom sized ad. This PR makes it possible get an ad with custom width + adaptive height.

### Usage
`<Banner size="300x" unitId="***" /> `
``<Banner size={`${SCREEN_WIDTH - MARGIN}x`} unitId="***" /> ``

### DIFF
--- a/RNGoogleMobileAdsExample/App.tsx
+++ b/RNGoogleMobileAdsExample/App.tsx
@@ -174,14 +174,16 @@ class InterstitialTest implements Test {
 }
 
 class BannerTest implements Test {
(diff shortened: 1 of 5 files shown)


Yassdrk:
hello,
why it is deleted ?

dylancom:
Wasn't so happy about this solution. Ideally it should take the width of the surrounding View.

PR #298, opened 2022-12-27
refactor: keep fluid size exclusive to GAMBanner

dylancom:
### Description
The FLUID size is exclusive to Google Ad Manager banners but it's currently part of the general BannerAdSize enum, which is confusing when using a regular banner.

As you can't extend enums, a const is used for GAMBannerAdSize (defined as const).
See: https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums


src/BannerAdSize.ts
```diff
@@ -58,13 +58,17 @@ export enum BannerAdSize {
    */
   INLINE_ADAPTIVE_BANNER = 'INLINE_ADAPTIVE_BANNER',
 
-  /**
-   * A dynamically sized banner that matches its parent's width and expands/contracts its height to match the ad's content after loading completes.
-   */
-  FLUID = 'FLUID',
```
mikehardy, line 64: I think this change is good and is the right thing, but this *is* a breaking change. A good one, but breaking yes? That just means when I merge unless the commit itself is marked breaking (ideally with the helpful instructions on how to handle it so it ends up auto-included in changelog) then it should be a merge commit (vs a rebase merge) with message:

`BREAKING: GAM-specific type FLUID moved to new const GAMBannerAdSize`

...and that should trigger both the semver major + be adequate to help folks adapt?

mikehardy (approved):
I like it - please merge + release and see if my comment on exactly how to merge is sufficient to trigger the semver major that I *think* is proper for this one?

PR #305, opened 2023-01-04
fix(banner, android): support props for multiple view instances

aidan-24com:
### Description

The Java class for `ReactNativeGoogleMobileAdsBannerAdViewManager` is a _single_ instance. It instantiates new `reactViewGroup` instances for _each_ JSX element that is rendered.

The Objective-C class `RNGoogleMobileAdsBannerViewManager` is instantiated for each JSX element.

This PR is to address the Java discrepancy, because when rendering multiple banners (on Android) the latest set props are used for the `reactViewGroup` instances.

### Related issues

I didn't file an issue for it.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Render multiple Banners in a list (on android) with different props for each ad slot. The received ads will now render correctly with their respective options.

Previously, due to the asynchronous nature of the bridge, ad slots that where already instantiated would start to use the latest set props.

---

:fire:


aidan-24com:
apologies, I'll run the linter and update the PR shortly

aidan-24com:
pinging a maintainer as requested above

@mikehardy when you get a chance 👍

dylancom:
Could you explain why the adlistener on oldAdView is needed?

aidan-24com:
Apologies, the oldAdView adlisteners code was basically a straight copy of the code we had running on an old version of this package.

After testing it again, the oldAdView adlisteners are not required.

Initially there was an issue that the listeners for ad views that had been removed were triggering. This appears to no longer be the case, I have updated the PR.

dylancom:
Thanks and is it necessary to move the oldAdView.destroy() line?

dylancom:
I did a bit of research and found that we can extend the ReactViewGroup to avoid having to work with maps.
That way we can directly set the props on the extended ReactViewGroup instance, which seems a neater approach to me.

See: https://github.com/sbugert/react-native-admob/blob/master/android/src/main/java/com/sbugert/rnadmob/RNAdMobBannerViewManager.java#L28

And: https://github.com/sbugert/react-native-admob/blob/master/android/src/main/java/com/sbugert/rnadmob/RNAdMobBannerViewManager.java#L220

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsBannerAdViewManager.java
```diff
@@ -55,12 +56,12 @@
   private final String EVENT_APP_EVENT = "onAppEvent";
   private final int COMMAND_ID_RECORD_MANUAL_IMPRESSION = 1;
 
-  private AdRequest request;
-  private List<AdSize> sizes;
-  private String unitId;
-  private Boolean manualImpressionsEnabled;
-  private boolean propsChanged;
-  private boolean isFluid;
+  private final HashMap<Integer, AdRequest> requestMap = new HashMap<>();
```
dylancom, line 59: If we create a ReactAdView class that extends ReactViewGroup, we don't need maps to keep track of everything and set those props directly on the instance. See: https://github.com/sbugert/react-native-admob/blob/master/android/src/main/java/com/sbugert/rnadmob/RNAdMobBannerViewManager.java#L28

dylancom (changes requested):

dylancom:
@aidan-24com is the requested change something you'd have time to do, or should I make a PR?

aidan-24com:
@dylancom I will attempt to get it in this week, since it is on my task list.

thanks for the offer. I might reach out, if things get too hectic

aidan-24com:
@dylancom when you have a moment, please take a look at the `ReactNativeAdView` implementation.

thanks

dylancom:
@aidan-24com was just testing it out on a real device.
Thanks a lot for bringing this up and your work! 🚀

dylancom (approved):

mikehardy:
Really cool - thanks @aidan-24com ! And also thanks to @dylancom for shepherding, really really appreciated

PR #338, opened 2023-03-02
docs(ios): add advice on requesting app tracking transparency

dylancom:
### Description

All iOS apps should request the app tracking transparency, so make this part of the installation flow.


mthomas100:
Apologies if this is covered elsewhere - is the ATT prompt a requirement when using this sdk?

PR #389, opened 2023-05-26
feat(catalyst): support building on catalyst, though ads will not work

### LINKED ISSUE #354: [🐛] The library breaks builds for Mac Catalyst
### What happened?

I've tried to build an existing Android/iOS RN app using **rn-google-mobile-ads** for Mac Catalyst target, and the build fails, because Google's Mobile Ads SDK does not support Mac Catalyst. However, I checked that if I comment out everything inside this library's native code, beside the high-level interface exposed to RN's JS layer, and also comment out the links to Mobile Ads SDK in this library's `.podspec` file &mdash; then Mac Catalyst build works fine (well, after fixing a bunch of minor issues in my other dependencies), and my app works fine at the runtime, at the first glance. I mean, it obviously does not show any ads, but it does not crash, beside no changes in my code, thus still doing all calls to **rn-google-mobile-ads** the same way it does on other platforms. It also still shows fallback ads (in my app I use banners, and I use interstitials, so normally if a banner or interstitial is not loaded the app shows a dummy ad in that place, inviting users to subscribe, to remove ads, and giving them links to the subscriptions page) &mdash; thus, effectively, it works quite good for me.

Hence, my suggestion is to add a bunch of `#ifndef TARGET_OS_MACCATALYST` pragmas into ios code, and whatever is the equivalent into `.podspec`, to ensure that when build for Mac Catalyst the library does not break the build, and does not crash at runtime, just does nothing when its functions are called.

What do you think about it @mikehardy ? Am I missing an alternative way to conditionally include / exclude RN library into an app, depending on the build target?

### Platforms

Only on iOS

### React Native Info

```Shell
irrelevant
```

### Are your using Typescript?

- [ ] My project is using Typescript

### package.json

```JSON
irrelevant
```

### app.json

```JSON
irrelevant
```

### ios/Podfile

_No response_

### android/build.gradle

_No response_

### android/app/build.gradle

_No response_

### android/settings.gradle

_No response_

### AndroidManifest.xml

_No response_

birdofpreyru:
### Description

Adds a bunch of pre-processor directives all around the codebase to ensure the library does not break Mac Catalyst builds, and all its APIs just do nothing on that platform.

### Related issues

Fixes #354 

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

I've tested it with a real project of mine (still can't make the example app in the repo work for me):
- It works as before when I am building my project with these changes.
- It does not break Mac Catalyst builds anymore, and my project works fine on my machine, with no additional code changes it just does not show any ads there. The only thing, for this to work the iOS pods should be installed like (`MAC_CATALYST=1 pod install` &mdash; I wasn't able to find a better way to conditionally detect target platform inside podspec, to exclude Google Ads SDK pod for Mac Catalyst scenario).

---

### DIFF
--- a/RNGoogleMobileAds.podspec
+++ b/RNGoogleMobileAds.podspec
@@ -30,7 +30,9 @@ Pod::Spec.new do |s|
     google_ump_sdk_version = $RNGoogleUmpSDKVersion
   end
 
+  if !ENV['MAC_CATALYST']
   s.dependency          'GoogleUserMessagingPlatform', google_ump_sdk_version
+  end
 
   if defined?($RNGoogleMobileAdsSDKVersion)
     Pod::UI.puts "#{s.name}: Using user specified Google Mobile-Ads SDK version '#{$RNGoogleMobileAdsSDKVersion}'"
@@ -38,7 +40,9 @@ Pod::Spec.new do |s|
   end
 
   # AdMob dependencies
+  if !ENV['MAC_CATALYST']
   s.dependency          'Google-Mobile-Ads-SDK', google_mobile_ads_sdk_version
+  end
 
   if defined?($RNGoogleMobileAdsAsStaticFramework)
     Pod::UI.puts "#{s.name}: Using overridden static_framework value of '#{$RNGoogleMobileAdsAsStaticFramework}'"
--- a/ios/RNGoogleMobileAds/RNGoogleMobileAdsAppOpenModule.m
+++ b/ios/RNGoogleMobileAds/RNGoogleMobileAdsAppOpenModule.m
@@ -15,6 +15,8 @@
  *
  */
 
+#if !TARGET_OS_MACCATALYST
+
 #import <React/RCTBridgeModule.h>
 
 @interface RCT_EXTERN_MODULE (RNGoogleMobileAdsAppOpenModule, NSObject)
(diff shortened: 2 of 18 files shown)


dylancom:
Hmm is this something many people benefit from or more for a fork / local patch (https://www.npmjs.com/package/patch-package)?

birdofpreyru:
@dylancom well... there are somewhat more than a few Mac devices on this planet, and RN allows to mostly seamlessly build the same app for both iOS and Mac, thus I'd say my desire that this library out of the box does not break Mac Catalyst builds isn't something niche / exotic, and some more ppl should find it helpful.

dylancom:
Alright, personally I find all the extra conditionals needed within the exported methods a bit error prone.
But leaving this up to @mikehardy who seems to have more experience with this.

mikehardy:
This is - as far as I know - still the best + really only way to support macCatalyst in a practical way. In practice the conditionals have not been painful to maintain in the other projects where I've merged (or implemented myself!) similar PRs.

Stated slightly differently, I definitely am sensitive to maintenance overhead and this seems like it might be a burden but in practice it has never been one over a few years, so I'm okay with it

Appear to have a lint issue though, a local run of `lint:ios:fix` should fix it, I'll do that and re-push

mikehardy (approved):
This should pass CI and looks good to me

PR #433, opened 2023-08-12
feat: return updated consent status from privacy options form

DoctorJohn:
### Description

The user can update their consent choices from the privacy options form, just like with the normal consent form (tbh I think they are the same). After the privacy options form is dismissed by the user we're most likely interested in the updated consent.

This PR lets the `showPrivacyOptionsForm` return the updated consent status just like the `showForm` method does.

Even the implementation is the same since both `showForm` and `showPrivacyOptionsForm` make use of the `ConsentForm.OnConsentFormDismissedListener` interface whose [documentation](https://developers.google.com/interactive-media-ads/ump/android/api/reference/com/google/android/ump/ConsentForm.OnConsentFormDismissedListener.html) recommends calling `ConsentInformation.getConsentStatus` to get the updated consent status.

I'm not sure whether we should consider this a breaking change. Technically it is because we change the return type. Before the method returned a constant string which hopefully none actually used in any way in their code.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [x] Yes
  - [ ] No

### Test Plan

Tried it with patch-package. Waiting for CI to turn green.

---

### DIFF
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsConsentModule.java
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsConsentModule.java
@@ -193,7 +193,10 @@ public class ReactNativeGoogleMobileAdsConsentModule extends ReactNativeModule {
                       rejectPromiseWithCodeAndMessage(
                           promise, "privacy-options-form-error", formError.getMessage());
                     } else {
-                      promise.resolve("Privacy options form presented successfully.");
+                      WritableMap resultMap = Arguments.createMap();
+                      resultMap.putString(
+                          "status", getConsentStatusString(consentInformation.getConsentStatus()));
+                      promise.resolve(resultMap);
                     }
                   }));
     } catch (Exception e) {
--- a/ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
+++ b/ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
@@ -165,7 +165,11 @@ RCT_EXPORT_METHOD(showPrivacyOptionsForm
                                                                formError.localizedDescription,
                                                          } mutableCopy]];
                                   } else {
-                                    resolve(@"Privacy options form presented successfully.");
+                                    resolve(@{
+                                      @"status" : [self
+                                          getConsentStatusString:UMPConsentInformation
+                                                                     .sharedInstance.consentStatus],
+                                    });
                                   }
                                 }];
 #endif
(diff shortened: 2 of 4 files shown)


dylancom:
He @DoctorJohn, thanks for your PR! I am doubting a bit whether it would be more neat to just expose a method "getConsentStatusString", just like we have "getTCString". And so stay closer to the native implementation.

DoctorJohn:
Hey @dylancom, if we want to stay closer to the native API I would propose a `getConsentInformation` method. The state of the `ConsentInformation` object is always available on the native side but only through `requestInfoUpdate` on the JS side.

The `ConsentInformation` object may update when `showForm` or `showPrivacyOptionsForm` are called. Currently there is no way to know what changed in this case without calling `requestInfoUpdate` again.

I don't think calling `requestInfoUpdate` more than once is even intended. I've been doing so for over a year now but just recently realized doing so has unexpected sideeffects (in a few occasions calling `requestInfoUpdate` reset the consent obtain from users).

Side note:

I just worked on a followup PR that would make both `showForm` and `showPrivacyOptionsForm` have the same return type as `requestInfoUpdate`, since whenever I call either of these methods I want to synchronize the "native" state with my apps state right afterwards. Pausing on that for now, since it would not match the native API.

DoctorJohn:
I could update this PR to add a `getConsentInformation` method. At least for me a `getConsentStatusString` would not be enough since I also want to get the updated `privacyOptionsRequirementStatus` and potentially the updated `canRequestAds`.

(I know this PR originally was only concerned with the updated consent status, but that was an oversight and the reason I started working on the followup PR I mentionend in the previous comment).

dylancom:
Just had a look at the native code (haven't been involved with most part of our UMP port btw) and you should be for example able to get the status of `UMPConsentInformation.sharedInstance.canRequestAds` without calling requestInfoUpdate multiple times.
https://developers.google.com/admob/ios/privacy#request-ads

DoctorJohn:
> Just had a look at the native code (haven't been involved with most part of our UMP port btw) and you should be for example able to get the status of `UMPConsentInformation.sharedInstance.canRequestAds` without calling requestInfoUpdate multiple times. https://developers.google.com/admob/ios/privacy#request-ads

Yes, absolutely! I meant from the JavaScript side though. The only JS method we have returning `canRequestAds` is `requestInfoUpdate`. So after showing a form I would have to call the `requestInfoUpdate` JS method again to see whether `canRequestAds` changed.

DoctorJohn:
Hey @dylancom, I created another PR with an updated description to hopefully make it clear what I'm trying to achieve.

PR #434, opened 2023-08-13
feat: return consent info whenever it changes

DoctorJohn:
### Description

UMP SDK version 2.1.0 exposes a few new methods and values which are not fully accessible though our wrapper yet. This PR proposes the following changes:

1. The existing `showForm` method currently only returns the updated value of the `getConsentStatus` method. However, calling `showForm` also updates the return value of `canRequestAds` (this is new in version 2.1.0), so I propose we let `showForm` return the updated result of `canRequestAds` as well.

2. The recently added `showPrivacyOptionsForm` method is essentially the same as the `showForm` method. It also updates the return values of `getConsentStatus` and `canRequestAds`. So I propose calling it should return exactly the same result type as the `showForm` method.

3. The upstream SDK unfortunately does not clearify whether showing forms only updates the return values of `getConsentStatus` and `canRequestAds` or whether `getPrivacyOptionsRequirementStatus` and `isConsentFormAvailable` may be affected as well. So I propose we just return these values as well and use one uniform return type for all methods updating the internal `ConsentInformation` object. These methods are: `showForm`, `showPrivacyOptionsForm` and `requestInfoUpdate`.

Alternative:

@dylancom mentionend in #433 that we might want to stay closed to the SDKs API. In this case we would have to change the return types of `showForm`, `showPrivacyOptionsForm` and `requestInfoUpdate` to `Promise<void>` and introduce a new method returning the state of the internal `ConsentInformation` object. I can create a PR for that as well, but IMHO this approach is more practical, it's also in line with our existing API.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

Ran tests and the example app.

---

### DIFF
--- a/ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
+++ b/ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
@@ -70,6 +70,20 @@ RCT_EXPORT_MODULE();
 }
 #endif
 
+#if !TARGET_OS_MACCATALYST
+- (NSDictionary *)getConsentInformation {
+  return @{
+    @"status" : [self getConsentStatusString:UMPConsentInformation.sharedInstance.consentStatus],
+    @"canRequestAds" : @(UMPConsentInformation.sharedInstance.canRequestAds),
+    @"privacyOptionsRequirementStatus" :
+        [self getPrivacyOptionsRequirementStatusString:UMPConsentInformation.sharedInstance
+                                                           .privacyOptionsRequirementStatus],
+    @"isConsentFormAvailable" :
+        @(UMPConsentInformation.sharedInstance.formStatus == UMPFormStatusAvailable)
+  };
+}
+#endif
+
 RCT_EXPORT_METHOD(requestInfoUpdate
                   : (NSDictionary *)options
                   : (RCTPromiseResolveBlock)resolve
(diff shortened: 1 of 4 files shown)


ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
```diff
@@ -125,24 +126,20 @@ - (NSString *)getPrivacyOptionsRequirementStatusString:
...
+                    completionHandler:^(NSError *_Nullable dismissError) {
+                      if (dismissError) {
+                        [RNSharedUtils
+                            rejectPromiseWithUserInfo:reject
+                                             userInfo:[@{
+                                               @"code" : @"consent-form-error",
+                                               @"message" : dismissError.localizedDescription,
+                                             } mutableCopy]];
+                      } else {
+                        resolve([self getConsentInformation])
```
dylancom, line 140: error: expected ';' after expression resolve([self getConsentInformation]).

ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
```diff
@@ -165,7 +162,7 @@ - (NSString *)getPrivacyOptionsRequirementStatusString:
                                                                formError.localizedDescription,
                                                          } mutableCopy]];
                                   } else {
-                                    resolve(@"Privacy options form presented successfully.");
+                                    resolve([self getConsentInformation])
```
dylancom, line 165: error: expected ';' after expression resolve([self getConsentInformation]).

dylancom (changes requested):
error: expected ';' after expression resolve([self getConsentInformation])

dylancom:
I agree that this is more practical! Thanks for the PR.

ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
DoctorJohn: Looks like I actually one ran the Android tests :O Just fixed both.

PR #516, opened 2024-01-02
feat: add method to check if gdpr applies

dylancom:
### Description

Add a method to expose the IABTCF_gdprApplies value.
So you can determine if GDPR rules applies in the current context.


mikehardy:
Nice!

PR #522, opened 2024-01-10
feat: add method to check if consent for purpose 1 was given

dylancom:
### Description

This feature makes it possible to only show the ATT alert if consent (for Purpose 1) was given.
As advised by: https://developers.google.com/admob/ios/privacy/gdpr#frequently_asked_questions

See also: https://developers.google.com/admob/ios/privacy/gdpr#how_to_read_consent_choices


mikehardy:
Overlap/duplication here with https://github.com/zoontek/react-native-permissions but...does seem to be a vital part of this module as well, seems okay

dylancom:
@mikehardy I'm not sure what do you mean with overlap/duplication?
This PR doesn't add the code to do the ATT request, this only adds code so you can check if "Consent Purpose 1" was given by the user.
Which then let's you decide if you are going to show the ATT alert at all. (As per Google's advice you don't need to show the ATT alert if consent was denied in the Google UMP form.

PR #531, opened 2024-01-25
feat(ios): minor correction for macOS builds

birdofpreyru:
### Description

Correct the latest lib version to pass the build for macOS (Catalyst).

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
+++ b/ios/RNGoogleMobileAds/RNGoogleMobileAdsConsentModule.m
@@ -200,7 +200,9 @@ RCT_EXPORT_METHOD(reset) {
 RCT_EXPORT_METHOD(getConsentInfo
                   : (RCTPromiseResolveBlock)resolve
                   : (RCTPromiseRejectBlock)reject) {
+#if !TARGET_OS_MACCATALYST
   resolve([self getConsentInformation]);
+#endif
 }
 
 RCT_EXPORT_METHOD(getTCString : (RCTPromiseResolveBlock)resolve : (RCTPromiseRejectBlock)reject) {


mikehardy (approved):
I don't see any reason not to trust you on this one, and this is the typical way to make catalyst builds work (I'm working through the same thing in other libs at the moment too...)

mikehardy (changes requested):
Oh wait - if you do this, *it never resolves*. I think you need to resolve null and make sure that the API is documented as returning null sometimes ?

birdofpreyru:
> Oh wait - if you do this, it _never resolves_.

Yeap, this way it never resolves.

> I think you need to resolve null and make sure that the API is documented as returning null sometimes ?

Not sure... for me never resolving does not cause any issues &mdash; it indefinitely blocks some ads-related logic, but at least in my apps it does not block anything non-ads related, thus its fine. In anyway, I guess similar macOS guards we added in past also cause other functions to never resolve on macOS, thus if you wanna go this way &mdash; resolving to null or something else, then we'll need to revise the rest of the library to ensure other functions behave similarly on Catalyst.

mikehardy:
I think not resolving is a major negative - I have had bugs logged against it in react-native-firebase and at some point we'll have to fix it, but it will be after we have forgotten what was happening and thus take longer than just doing it now while we are thinking about it.

If there are other methods that are also not resolving because of IFDEFs for platforms, fixing those would be good as well, so I'd like to do so

I recognize I'm asking for a change that you indicate doesn't matter in your app, but...I also think resolving in these situations across the whole library is really just a few minutes work so ... please 🙏 😄

birdofpreyru:
Not stale, but no ETA when I get to do it :)

birdofpreyru:
Not stale, but no ETA when I get to do it :)

PR #579, opened 2024-05-20
feat: add method to reload banner ad

dylancom:
### Description

As banner ads can become blank when an app was in background, the advise is to load a new ad upon returning to foreground. Therefore I added a method to make this possible.


ugurdalkiran:
It's a great feature. Thank you very much. 🥳

DoctorJohn:
Great changes @dylancom and straight forward code. I'll give them a quick test in my apps.

dylancom:
Thanks! I slightly refactored it so there is some code reusage and called it "load".

PR #604, opened 2024-07-10
refactor: modified the ios_config script to use single quotes instead…

ccorneliusHsv:
… of backslashes

### Description

A recent change to ios_config.sh introduced a line with backslashes. Backslashes are known to cause issues with React-Native-Version. See https://github.com/stovmascript/react-native-version/issues/327. This is because of a known issue with pbxproj-dom https://github.com/NativeScript/pbxproj-dom/issues/9. It does not appear that this issue is going to be addressed in those projects. My proposed solution is switching from a backslash to single quotes around the special character (https://linuxsimply.com/bash-scripting-tutorial/string/basics/escape-string/).  This is a refactor where no behavior changed. 

I created a patch to accomplish this in my project, but I thought since it might be affecting others, I would submit a PR. 
### Release Summary

swapped from backslashes to single quotes to play nice with react-native-version
### Checklist

  - [X] Yes
  - [ ] `Android`
  - [X] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

### Test Plan

---

🔥

### DIFF
--- a/ios_config.sh
+++ b/ios_config.sh
@@ -93,7 +93,7 @@ while true; do
 
   echo "info:      ($_CURRENT_LOOKUPS of $_MAX_LOOKUPS) Searching in '$_CURRENT_SEARCH_DIR' for a ${_JSON_FILE_NAME}/${_JS_APP_CONFIG_FILE_NAME} file."
 
-  _SEARCH_RESULT=$(find "$_CURRENT_SEARCH_DIR" -maxdepth 2 \( -name ${_JSON_FILE_NAME} -o -name ${_JS_APP_CONFIG_FILE_NAME} \) -print | /usr/bin/head -n 1)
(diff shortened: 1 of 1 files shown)


dylancom:
Hi, thanks for the PR. But I think a patch is sufficient in this case supporting a library that had it's last release in march 2020.

mikehardy (approved):
I dunno, I can't promise we won't accidentally break it again as there's no testing around the case, but it's kind of a neat shell scripting trick, I'm okay with it though yes @dylancom it's a bit cheese to be doing a workaround for a transitive transitive that's out of date 😆 . 

I read the bash link you included (thank you) and it was good to have a refresher on that anyway, so I'm +1 on this tiny patch as it forced me to refresh some knowledge that was fading anyway 🤷  works for me

PR #606, opened 2024-07-12
chore: remove redundant app.config.js support

DoctorJohn:
### Description

This PR removes the redundant and deprecated handling of `app.config.js` files from the `ios_config.sh` and `build.gradle` scripts.

Now that `app.config.js` support is handled by the Expo config plugin, the `ios_config.sh` and `build.gradle` scripts no longer need to consider `app.config.js` at all. Parts of the code added in #517 can be removed again to make the scripts easier to maintain. That's exactly what this PR does. It's simply reverting all changes to the relevant files from said PR. I double checked that this does only revert the `app.config.js` support and nothing else (e.g. fixes or other refactorings).

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- Ran our tests
- Made sure the example app still works

### DIFF
--- a/ios_config.sh
+++ b/ios_config.sh
@@ -28,13 +28,6 @@
 
 set -e
 
-if [[ -f "$PODS_ROOT/../.xcode.env" ]]; then
-  source "$PODS_ROOT/../.xcode.env"
-fi
-if [[ -f "$PODS_ROOT/../.xcode.env.local" ]]; then
-  source "$PODS_ROOT/../.xcode.env.local"
-fi
-
 _MAX_LOOKUPS=2;
 _SEARCH_RESULT=''
 _RN_ROOT_EXISTS=''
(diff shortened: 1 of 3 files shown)


ios_config.sh
```diff
@@ -28,28 +28,18 @@
...
 _JSON_ROOT="'react-native-google-mobile-ads'"
 _JSON_FILE_NAME='app.json'
-_JS_APP_CONFIG_FILE_NAME='app.config.js'
 _JSON_OUTPUT_BASE64='e30=' # { }
 _CURRENT_SEARCH_DIR=${PROJECT_DIR}
 _PLIST_BUDDY=/usr/libexec/PlistBuddy
 _TARGET_PLIST="${BUILT_PRODUCTS_DIR}/${INFOPLIST_PATH}"
 _DSYM_PLIST="${DWARF_DSYM_FOLDER_PATH}/${DWARF_DSYM_FILE_NAME}/Contents/Info.plist"
-_IS_CONFIG_JS=false
-_PACKAGE_JSON_NAME='package.json'
```
dylancom, line 52: _PACKAGE_JSON_NAME is used to bail out if using expo. So this breaks it.

dylancom (changes requested):

ios_config.sh
DoctorJohn: Thanks Dylan, I didn't notice. Should have tested on Expo too.
Fixed it.

dylancom:
Good catch👍

PR #615, opened 2024-07-25
chore: remove unused file

DoctorJohn:
### Description

I always stumble across this file that I believe was commited by mistake since it does not contain real test, does not run as part of the test suite and is not referenced anywhere.

### Test Plan

- Tests still pass
- Checked that file is not referenced anywhere


dylancom:
I'm sure it hasn't been included by mistake, it's just currently unused.
Seems we have to add this to our tsconfig for type tests during compilation.
See: https://github.com/invertase/react-native-firebase/blob/c2c457608c865f6d25e4b9b6689cbc00721de194/tsconfig.json#L2C3-L2C69

(The origin where we extracted this package from)

DoctorJohn:
Closed in favor of #619

Issue #620, opened 2024-07-31
[🐛] No signature of method: java.lang.Boolean.getStringValue() under android v14.1.0+


xdmx:
### What happened?

I've been using this library for a while (thank you!) and recently migrated to the expo config, which worked well under 14.0.1, then I upgraded to 14.2.1 and I've started getting the compilation under android failing with:
```
> Configure project :react-native-google-mobile-ads
:react-native-google-mobile-ads package.json found at /Projects/app/node_modules/react-native-google-mobile-ads/package.json
:app (dev) package.json found at /Projects/app/package.json

FAILURE: Build failed with an exception.

* Where:
Build file '/Projects/app/node_modules/react-native-google-mobile-ads/android/build.gradle' line: 85

* What went wrong:
A problem occurred evaluating project ':react-native-google-mobile-ads'.
> No signature of method: java.lang.Boolean.getStringValue() is applicable for argument types: (String, String) values: [android_app_id, ]

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

```

I've tried older versions and it also happened with 14.1.0 and 14.2.0. 14.0.1 is the latest that works, without doing any change in the config or code other than changing the version in the package.json. It both fails in local and EAS compilations

Ios is working (compiling through EAS, I cannot do it locally)

### Platforms

Only on Android

### React Native Info

```Shell
System:
  OS: Linux 6.10 Arch Linux
  CPU: (16) x64 AMD Ryzen 7 PRO 4750U with Radeon Graphics
  Memory: 8.60 GB / 29.10 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 20.14.0
    path: ~/.nodenv/versions/20.14.0/bin/node
  Yarn:
    version: 1.22.22
    path: /usr/bin/yarn
  npm:
    version: 10.7.0
    path: ~/.nodenv/versions/20.14.0/bin/npm
  Watchman:
    version: 20240414.112832.0
    path: /usr/bin/watchman
SDKs:
  Android SDK: Not Found
IDEs:
  Android Studio: AI-241.18034.62.2411.12071903
Languages:
  Java:
    version: 17.0.12
    path: /usr/bin/javac
  Ruby:
    version: 3.3.2
    path: ~/.rbenv/shims/ruby
npmPackages:
[...]

PR #625, opened 2024-08-07
docs: clarify app.config.*s files are supported

DoctorJohn:
### Description

This PR clarifies that `app.config.js` and `app.config.ts` files are still supported (just like `app.json` files).

~This PR complements the `app.json` config examples with `app.config.js` and `app.config.ts` examples.~

~I was hoping we would not have to do this since it's somewhar cluttering the docs, but I tried to compensate that by using docs.page accordion components. Also tried using nested tabs, but they don't appear to be supported (i.e. they don't render correctly).~

### Related issues

- #620
  Hope this helps with issues like this, where some people got the impression that `app.config.*s` support was dropped

### Test Plan

- Validate the changes via the [docs preview](https://docs.page/DoctorJohn/react-native-google-ads~add-app-config-examples)

### DIFF
--- a/docs/european-user-consent.mdx
+++ b/docs/european-user-consent.mdx
@@ -34,9 +34,10 @@ For Android, add the following rule into `android/app/proguard-rules.pro`:
 </TabItem>
 <TabItem value="expo">
 
-Add the `extraProguardRules` property to `app.json` file as described in the [Expo documentation](https://docs.expo.dev/versions/latest/sdk/build-properties/#pluginconfigtypeandroid):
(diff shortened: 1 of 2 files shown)


dylancom:
Hmm I am a bit unsure if we need to add all this.

DoctorJohn:
> Hmm I am a bit unsure if we need to add all this.

After sleeping another night over it I think the same. It shouldn't be our job to teach users about three different ways to create an expo config file. I'll remove the accordions and app.config.*s examples again and instead only mention the support for app.config.*s and maybe refer users to the expo docs. Does that sound good @dylancom ?

dylancom:
Sounds good @DoctorJohn !

Issue #631, opened 2024-08-23
[🐛] Jest Setup file is deleting `TurboModuleRegistry.get`, breaking test suites


rodperottoni:
### What happened?

In my Jest Config file, I am adding `./node_modules/react-native-google-mobile-ads/jest.setup.ts` to my `setupFiles` config. This always worked for me until I decided to begin upgrading my libraries. That's when I started getting errors such as this in my test suite:

<img width="604" alt="Screenshot 2024-08-23 at 3 10 21 PM" src="https://github.com/user-attachments/assets/c9b3272f-068c-484e-8c25-d74320d5a1f9">

It took me a long time to find out the culprit, and apparently it's this file under `react-native-google-mobile-adds`. Notice how it assigns `getEnforcing` but not a `get` to the `TurboModulesRegistry` library:

```import * as ReactNative from 'react-native';

jest.doMock('react-native', () => {
  return Object.setPrototypeOf(
    {
      Platform: {
        OS: 'android',
        select: () => {},
      },
      NativeModules: {
        ...ReactNative.NativeModules,
        RNAppModule: {
          addListener: jest.fn(),
          removeListeners: jest.fn(),
          eventsAddListener: jest.fn(),
          eventsNotifyReady: jest.fn(),
        },
        RNGoogleMobileAdsModule: {
          addListener: jest.fn(),
          removeListeners: jest.fn(),
          eventsAddListener: jest.fn(),
          eventsNotifyReady: jest.fn(),
        },
        RNGoogleMobileAdsInterstitialModule: {
          interstitialLoad: jest.fn(),
        },
        RNGoogleMobileAdsRewardedModule: {},
        RNGoogleMobileAdsConsentModule: {},
      },
      TurboModuleRegistry: {
        getEnforcing: () => {
          return {
            initialize: jest.fn(),
            setRequestConfiguration: jest.fn(),
            openAdInspector: jest.fn(),
            openDebugMenu: jest.fn(),
            setAppVolume: jest.fn(),
            setAppMuted: jest.fn(),
          };
        },
      },
    },
    ReactNative,
  );
});
```

I think we need to better document how to setup `react-native-google-mobile-ads` with Jest, as well as fixing this mock for future users.

### Platforms

Android and iOS

### React Native Info

```Shell
Not applicable.
```

### Are your using Typescript?

- [X] My project is using Typescript

### package.json

```JSON
Not applicable.
```

### app.json

```JSON
Not applicable.
```

[...]

PR #619, opened 2024-07-27
test: fix outdated and unused type tests

DoctorJohn:
### Description

This PR fixes that our compile time type tests were never evaluated. Furthermore, it also updates the tests to include all types that have been added since their last update.

### Related issues

- Fixes #615 

### Checklist

- I read the [Contributor Guide](https://github.com/invertase/react-native-google-mobile-ads/blob/main/CONTRIBUTING.md)
  and followed the process outlined there for submitting PRs.
  - [x] Yes
- My change supports the following platforms;
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
  - [ ] `e2e` tests added or updated in `__tests__e2e__`
  - [ ] `jest` tests added or updated in `__tests__`
- [ ] I have updated TypeScript types that are affected by my change.
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- Made sure `yarn tsc:compile` now passes
- Checked that all types exported from `src/index.ts` are covered

PR #643, opened 2024-10-07
fix: type tests were messing with type generation

DoctorJohn:
### Description

As reported in https://github.com/invertase/react-native-google-mobile-ads/issues/631#issuecomment-2397953899, PR #619 broke the generation of TS types shipped with the package.

The reason for this is the inclusion of the `type-test.ts` file in the `include` array in the `tsconfig.json` file. This was done so that the `tsc:compile` script used by the lint CI workflow would try to compile that file and report any issues with it. This was inspired by how it's done in the [react-native-firebase repo's tsconfig.json](https://github.com/invertase/react-native-firebase/blob/9c940857a3759f85fad2f8d995b83f7c08058d6b/tsconfig.json#L2).
However, unlike react-native-fiebase we ship those compiled files and the inclusion of the `type-test.ts` file is messing with the structure of the output files. Without the inclusion of the `type-test.ts` file, `yarn build` puts the types into `lib/typescript`. With the inclusion of the `type-test.ts` file, `yarn build` puts the type into `lib/typescript/src` - which is causing the reported issue.

This PR splits the `tsconfig.json` file into one used for compiling the types we ship and a `tsconfig.test.json` file used by the `tsc:compile` script for linting / testing the types.

### Test Plan

- `yarn build` outputs the expected types again
- `yarn tsc:compile` still passes
- `yarn tsc:compile` fails if any non-existing property is accessed from the `type-test.ts` file

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "lint:markdown:fix": "prettier --write \"docs/**/*.md[x]\"",
     "lint:report": "eslint --output-file=eslint-report.json --format=json src/ --ext .js,.jsx,.ts,.tsx",
     "lint:spellcheck": "spellchecker --quiet --files=\"docs/**/*.md\" --dictionaries=\"./.spellcheck.dict.txt\" --reports=\"spelling.json\" --plugins spell indefinite-article repeated-words syntax-mentions syntax-urls frontmatter",
-    "tsc:compile": "tsc --project . --noEmit",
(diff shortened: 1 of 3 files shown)


mikehardy (approved):
No good deed unpunished, they say
When I first read the description I was a bit worried the test-related tsconfig would be a big "don't repeat yourself" violation, but having it be a pure extension + no-emit and the type test file is about as clean as I can imagine it. Very nice

PR #654, opened 2024-11-15
feat: gather consent method to make user consent easier

dylancom:
### Description

While checking the official ios docs for changes I noticed that Google uses a helper method in their examples which I think can benefit this library too. Making it less cumbersome to show a consent form when necessary.

So you don't first have to call `requestInfoUpdate` and then `loadAndShowConsentFormIfRequired `.
But just `gatherConsent`.

See: https://github.com/googleads/googleads-mobile-ios-examples/blob/043de5a21d5b529ded1fb845f9f8082d130113f8/Objective-C/admob/BannerExample/BannerExample/GoogleMobileAdsConsentManager.m#L45

And: https://developers.google.com/admob/ios/privacy#request-ads

### DIFF
--- a/docs/european-user-consent.mdx
+++ b/docs/european-user-consent.mdx
@@ -198,22 +198,17 @@ import mobileAds, { AdsConsent, AdsConsentStatus } from 'react-native-google-mob
 
 let isMobileAdsStartCalled = false;
 
-// Request an update for the consent information.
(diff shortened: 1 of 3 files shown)


mikehardy (approved):
LGTM good spot

dylancom:
@ng-ha You're right. Thanks for mentioning it.

PR #662, opened 2024-11-26
feat(ios): full screen ads on new arch

dylancom:
### Description

Migrating full screen ads (**App Open, Interstitial, Rewarded Interstitial, Rewarded**) to make use of the new arch (old arch compatible).
I tested it in the example app on the new arch (default) and on the old arch by adding `ENV['RCT_NEW_ARCH_ENABLED'] = '0'` to the top of ios/Podfile.

Would be great if many people can battle test this. @DoctorJohn @birdofpreyru @TaylorDale ?


ios/RNGoogleMobileAds/RNGoogleMobileAdsAppOpenModule.h
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#import <Foundation/Foundation.h>
+
+#ifdef RCT_NEW_ARCH_ENABLED
+
+#import <RNGoogleMobileAdsSpec/RNGoogleMobileAdsSpec.h>
```
mikehardy, line 22: This port will be a regression for macCatalyst builds - all of the google ads stuff has to be ifdef'd out there so that it at least compiles in that mode. It won't *do* anything but it will at least compile

see https://github.com/invertase/react-native-google-mobile-ads/pull/389/files

adding back in the big / bulk ifdefs there is needed to preserve it

ios/RNGoogleMobileAds/RNGoogleMobileAdsAppOpenModule.mm
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#import "RNGoogleMobileAdsAppOpenModule.h"
```
mikehardy, line 18: I won't comment on every file everywhere but basically, before any RNGoogleMobile... or UMP... or GoogleMobile... anything symbol is touched, in each .h and .mm, have to do the macCatalyst ifdef, then endif way down at the end

mikehardy (commented):
looks pretty good to me unfortunately not-super-well-trained-new-architecture eyes

just the note about macCatalyst build ifdef-ing. You can test the build on that one by opening the xcworkspace in Xcode, temporarily adding your xcode dev team (may need to change the bundle id to something random for it to allow) adding macCatalyst to the possible build modes and setting the target to your mac. Then edit Podfile to turn on the react-native maccatalyst flag + pod install, finally hitting the build button

expressed in automated code using python `pbxproj` utility, looks like: https://github.com/mikehardy/rnfbdemo/blob/d5311b4b955fefc55b26d3ae47a94eb0cc5f27f3/make-demo.sh#L287-L306

mikehardy (approved):
the catalyst thing was the only thing I saw, if it is building and working for you in the example app on iOS and macCatalyst, works for me, and awesome, another huge chunk moved to new arch natively vs interop

dylancom:
I do get the error `Undefined symbol: _RNGoogleMobileAdsBannerViewCls` on Catalyst, which isn't related to this PR though... Might already not work in Catalyst?

PR #663, opened 2024-11-28
feat(banner): Support maximum height for Inline Adaptive banners

### LINKED ISSUE #541: [Feature Request] Support the max height option for inline adaptive banners
### What happened?

Inline adaptive banners are designed for scrolling content. If that content is a feed with fairly consistent cell sizes then the larger sizes of banner available can be terrible for the UX.

The native SDKs support a method to get an inline adaptive banner with a width and max height (rather than just a width) but we can't access this through the React Native SDK.

The standard banners don't work so well in a feed in practice because they always take up their fixed height while loading and the API doesn't provide any feedback on loading unless it fails.

It would be ideal if the `BannerAd` component could take an optional max height that if supplied for an inline adaptive banner would cause the alternative native SDK method to be used and restrict the height. :pray:

### Platforms

Android and iOS

### React Native Info

```Shell
All recent versions
```

### Are your using Typescript?

- [ ] My project is using Typescript

### package.json

```JSON
n/a
```

### app.json

```JSON
n/a
```

### ios/Podfile

_No response_

### android/build.gradle

_No response_

### android/app/build.gradle

_No response_

### android/settings.gradle

_No response_

### AndroidManifest.xml

_No response_

markwilcox:
### Description

Inline adaptive banners are the best variety to use on cell-based layouts like list views. They can also maximise revenue on scrollViews by taking advantage of more available inventory at different sizes. However they can grow very tall and ruin the UX in these situations.

The native iOS and Android SDKs support creating an inline adaptive banner with a maximum height to solve this issue. This PR is intended to add support for that behaviour from React Native.

### Related issues

Fixes #541 

### Release Summary

Add support for optional maxHeight property on BannerAd

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

I couldn't see any relevant tests to add to in e2e or __tests__. The new prop is optional and only has any effect on an inline adaptive banner. I went for forgiving - e.g. setting the prop on the wrong banner format is ignored, setting a prop smaller than the minimum supported by the native SDKs automatically increases it to the minimum size (32px).

I added a test in the example app and ensured it worked there on old and new architectures. Let me know if you want any other tests added, or to be more strict validating use of the prop (e.g. against minimum size, or in combination with other ad formats).

I'm not 100% on this implementation because of the way we need both the `sizes` prop and the `maxHeight` prop to set a size in the native ad request. As far as I know the order of calling these isn't guaranteed, so I'm storing the original sizes array, and recalculating the sizes if the maxHeight prop gets set on the native side. It works, but I wonder if it will ultimately create 2 requests for each ad shown and impact the show rate? If you can suggest a better implementation strategy without changing the API too much I'm willing to give it a go.

If you think this approach is OK then I'll update the docs.

:fire:
---


RNGoogleMobileAdsExample/App.tsx
```diff
@@ -1025,7 +1030,7 @@ TestRegistry.registerTest(new DebugMenuTest());
 const App = () => {
   return (
     <SafeAreaView>
-      <ScrollView contentInsetAdjustmentBehavior="automatic">
+      <ScrollView contentContainerStyle={{ paddingRight: 15 }} contentInsetAdjustmentBehavior="automatic">
```
markwilcox, line 1033: This was a quick hack because there are too many tests to fit on a screen already and they're all 100% pressable area, so you can't scroll the scroll view. This gives you some space on the right to scroll it.

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsBannerAdViewManager.java
```diff
@@ -135,11 +135,18 @@ public void setSizes(ReactNativeAdView reactViewGroup, ReadableArray value) {
       payload.putDouble("height", adSize.getHeight());
       sendEvent(reactViewGroup, EVENT_SIZE_CHANGE, payload);
     }
-
+    reactViewGroup.setSizesArray(value);
     reactViewGroup.setSizes(sizeList);
     reactViewGroup.setPropsChanged(true);
   }
 
+  @ReactProp(name = "maxAdHeight")
```
markwilcox, line 143: I wanted to call the native prop "maxHeight" but that ends up setting the native view height too, which isn't what you want in the case where the user sets a height lower than the minimum (32px) because you get a minimum height ad back and then the view crops it.

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsCommon.java
```diff
@@ -51,8 +52,16 @@ static AdSize getAdSizeForAdaptiveBanner(String preDefinedAdSize, ViewGroup reac
       DisplayMetrics outMetrics = new DisplayMetrics();
       display.getMetrics(outMetrics);
       int adWidth = (int) (outMetrics.widthPixels / outMetrics.density);
+      float maxAdHeight = ((ReactNativeAdView)reactViewGroup).getMaxAdHeight();
 
       if ("INLINE_ADAPTIVE_BANNER".equals(preDefinedAdSize)) {
+        if (maxAdHeight > 0) {
+          if (maxAdHeight < 32) {
```
markwilcox, line 59: How would you prefer a constant defined for this magic number?

ios/RNGoogleMobileAds/RNGoogleMobileAdsCommon.mm
```diff
@@ -206,6 +206,13 @@ + (GADAdSize)stringToAdSize:(NSString *)value {
     }
     CGFloat viewWidth = frame.size.width;
     if ([value isEqualToString:@"INLINE_ADAPTIVE_BANNER"]) {
+      if (maxHeight > 0) {
+        if (maxHeight < 32) {
```
markwilcox, line 210: Again here, how would you prefer a constant defined for the magic number?

dylancom:
Hi, thanks for the PR. I am not sure about the potentional double requests. Also the extra "sizeStrings" array condition isn't very clear.

My current thinking is that we could also adapt the current `size` prop and create some helper to provide a max height.
`<Banner size={getInlineAdaptiveBannerAdSize(100)} />` or width and height: `<Banner size={getInlineAdaptiveBannerAdSize(320, 100)} />`.
The helper method outputs a string that `stringToAdSize` will understand. So it could become something like: "INLINE_ADAPTIVE_BANNER_100".

The Flutter implementation for e.g. has this method: AdSize.getInlineAdaptiveBannerAdSize(int width, int maxHeight).

@mikehardy what do you think?

markwilcox:
> My current thinking is that we adapt the current `size` prop and create some helper to provide a max height: `<Banner size={getInlineAdaptiveBannerAdSize(320, 100)} />`. The helper method outputs a string that `stringToAdSize` (with small changes) will understand. So it could become something like: `"INLINE_ADAPTIVE_BANNER_300x100" || "INLINE_ADAPTIVE_BANNER_300_100"`.

Currently from React Native we can't set the width anywhere on the adaptive banners as it's hardcoded to use the screen width.

If I was going to change the API I'd change that too. Maybe keep the JS API but allow separate width and maxHeight to be passed, then make them all a single object with the size string that gets passed across the bridge (or set as a single prop in the new architecture)?

Going for conversion to strings on the JS side and then parsing those strings feels a bit messy. The alternative I guess would be to go for minimal change and only support a fixed set of heights. Probably revenue vs. UX impact the only sizes that will really matter for height restriction are 50px, 100px & 250px.

We'll probably experiment with sizes on the fork I've made in production (millions of monthly impressions) so I can answer some questions about double requests (ideally the native SDKs would cancel in-flight requests if you made another one immediately, but no idea if they do) / show rate, and eCPM impact at different sizes empirically.

mikehardy:
🤔 

> Currently from React Native we can't set the width anywhere on the adaptive banners as it's hardcoded to use the screen width.
>
> If I was going to change the API I'd change that too. Maybe keep the JS API but allow separate width and maxHeight to be passed, then make them all a single object with the size string that gets passed across the bridge (or set as a single prop in the new architecture)?

That was surprising to read for me. I don't use this package so I am a maybe-interesting combination of not that knowledgeable but also interested in helping the package do well. "maybe-interesting" because I have a fresh set of eyes for most of these things and even I understand that control over the UX in design is vital, and that means ability to constrain width and height, yes? Thus, surprising we don't permit real control here

Looking at Flutter implementation, it's in there - width and height are fundamental on the size supertype:

https://github.com/googleads/googleads-mobile-flutter/blob/6aa897e3dcc2b2d97f4df61192b015388ce876d6/packages/google_mobile_ads/android/src/main/java/io/flutter/plugins/googlemobileads/FlutterAdSize.java#L22-L25


...some types have extra stuff like orientation and what not, and they have a lot of machinery in there for factory constructor type things, but width and height are controllable

We...do not have that based on what I see, right?

https://github.com/invertase/react-native-google-mobile-ads/blob/2aa29f029461acf45472ed7d2683988ca577577f/src/BannerAdSize.ts#L19

So, I'd say this is a worthy API change as well. Backwards-compatible or not - if I understand things correctly it seems widthxheight control is critical and adding them would be "doing it right"

> Going for conversion to strings on the JS side and then parsing those strings feels a bit messy

Agreed there - that seems like the wrong way to go vs allowing width and height to be specifiable first-class and passed separately from JS to native

mikehardy:
Even the Flutter example for inline adaptive takes great care to determine the size (in their case, meaning width) *minus the gutters*. It's so important to control the width they show it in their "here is our most basic imagined use so you can just try it" example

dylancom:
> 
> Going for conversion to strings on the JS side and then parsing those strings feels a bit messy. The alternative I guess would be to go for minimal change and only support a fixed set of heights. Probably revenue vs. UX impact the only sizes that will really matter for height restriction are 50px, 100px & 250px.

I also agree that this feels messy. I don't know who invented this in this lib but it's what's already there. So that's where it came from:

```
Size: BannerAdSize | string;
_The size of the banner. Can be a predefined size via `BannerAdSize` or custom dimensions, e.g. `300x200`_
```

Anyway I find adjusting our current api and using the size prop neater than adding additional props.
See Flutter example:
        
```
AdSize size = AdSize.getCurrentOrientationInlineAdaptiveBannerAdSize(
        _adWidth.truncate());`

BannerAd(
      adUnitId: 'ca-app-pub-3940256099942544/9214589741',
      size: size,
      request: AdRequest()
)
```

dylancom:
> 
> Going for conversion to strings on the JS side and then parsing those strings feels a bit messy. The alternative I guess would be to go for minimal change and only support a fixed set of heights. Probably revenue vs. UX impact the only sizes that will really matter for height restriction are 50px, 100px & 250px.

I agree, that suggestion came from using what is already available in this lib (you can pass sizes such as "300x200")
which are old ideas now:

```
Size: BannerAdSize | string;
_The size of the banner. Can be a predefined size via `BannerAdSize` or custom dimensions, e.g. `300x200`_
```

So now is a great time to adjust our current api.
My personal preference goes to keep using the size prop we already have instead of adding additional props.
This is also how Flutter does it and I think we can take notes there.

Flutter example:
        
```
AdSize size = AdSize.getCurrentOrientationInlineAdaptiveBannerAdSize(
        _adWidth.truncate());`
        
// If you wish to limit the height of the banner, you can use the static method
// AdSize.getInlineAdaptiveBannerAdSize(int width, int maxHeight)

BannerAd(
      adUnitId: 'ca-app-pub-3940256099942544/9214589741',
      size: size,
      request: AdRequest()
)
```

markwilcox:
> So now is a great time to adjust our current api. My personal preference goes to keep using the size prop we already have instead of adding additional props. This is also how Flutter does it and I think we can take notes there.
> 
> Flutter example:
> 
> ```
> AdSize size = AdSize.getCurrentOrientationInlineAdaptiveBannerAdSize(
>         _adWidth.truncate());`
>         
> // If you wish to limit the height of the banner, you can use the static method
> // AdSize.getInlineAdaptiveBannerAdSize(int width, int maxHeight)
> 
> BannerAd(
>       adUnitId: 'ca-app-pub-3940256099942544/9214589741',
>       size: size,
>       request: AdRequest()
> )
> ```

The Flutter API is doing a lower level wrapper around the native SDK AdSize creation methods. To me that seems like an unnecessary extra round trip to native land for a trivial bit of config. Generating the value of a prop in native land and then passing it back again isn't really "the React Native way" in my experience.

We could replicate that without the extra round trip by calling a helper function that's pure JS and just constructs an object with the necessary params to pass across to native. Or we could just have separate (optional) props on the BannerAd itself that get converted to the appropriate object to pass to native. I'd be OK with going for either of those. I think separate optional props feels more React-like as an API, but the helper function is closer to the native SDK & Flutter APIs if consistency there is more helpful.

If there's a consensus on which one to go for I can update the PR. Might need a couple of weeks to prioritise it again.

dylancom:
Oke so my first preference went towards bundling all size related things in the "size" prop.
But I am fine with an extra `width` and `maxHeight` prop.

Thanks for contributing!

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsCommon.java
```diff
@@ -51,8 +52,16 @@ static AdSize getAdSizeForAdaptiveBanner(String preDefinedAdSize, ViewGroup reac
...
+            return AdSize.getInlineAdaptiveBannerAdSize(adWidth, 32);
+          } else {
+            return AdSize.getInlineAdaptiveBannerAdSize(adWidth, Math.round(maxAdHeight));
+          }
+        }
```
dylancom, line 64: I believe you can do something like:
```suggestion
    if (maxAdHeight > 0) {
        return AdSize.getInlineAdaptiveBannerAdSize(adWidth, Math.max(maxAdHeight, 32));
    }
    return AdSize.getCurrentOrientationInlineAdaptiveBannerAdSize(
        reactViewGroup.getContext(), adWidth);
```

dylancom (changes requested):

ios/RNGoogleMobileAds/RNGoogleMobileAdsBannerComponent.m
```diff
@@ -78,6 +83,14 @@ - (void)setSizes:(NSArray *)sizes {
   _propsChanged = true;
 }
 
+- (void)setMaxAdHeight:(CGFloat)maxAdHeight {
+  _maxAdHeight = maxAdHeight;
+  if (_sizeStrings != nil) {
```
dylancom, line 88: From my testing in the example app, this condition never becomes truthy. So do we really need this extra sizeStrings?

dylancom (changes requested):

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsBannerAdViewManager.java
```diff
@@ -135,11 +135,21 @@ public void setSizes(ReactNativeAdView reactViewGroup, ReadableArray value) {
...
+  public void setMaxAdHeight(ReactNativeAdView reactViewGroup, float value) {
+    reactViewGroup.setMaxAdHeight(value);
+    ReadableArray adSizes = reactViewGroup.getSizesArray();
+    if (adSizes != null) {
+      this.setSizes(reactViewGroup, adSizes);
+    }
```
dylancom, line 149: From my testing in the example app, this condition is never true. So we might not need the extra sizesArray.

dylancom (changes requested):

src/types/BannerAdProps.ts
```diff
@@ -48,6 +48,7 @@ export interface BannerAdProps {
    */
   size: BannerAdSize | string;
 
+  maxHeight?: number;
```
dylancom, line 51:
```suggestion
  /**
   * Limit inline adaptive banner height.
   * By default, inline adaptive banners instantiated without a maxHeight value have a maxHeight equal to the device height.
   */
  maxHeight?: number;
```

dylancom (changes requested):

dylancom:
If we can't be sure that maxHeight is set before sizes we might group them in BaseAd: sizeConfig={{ sizes, maxHeight }}. So the native side receives them as a bundle but the user can just use seperate size and maxHeight props on the <Banner /> component. Thanks @mikehardy for the idea.

ios/RNGoogleMobileAds/RNGoogleMobileAdsBannerComponent.m
markwilcox: Yes, I couldn't reproduce this on iOS or Android, but I was missing the same check on Android initially and running this with ~700k Android users in production, I was getting 10s of crashes a day.

My plan is to refactor this to pass an object with the size related props across to native in a single prop, while having separate props exposed on the JS side.

markwilcox:
> If we can't be sure that maxHeight is set before sizes we might group them in BaseAd: sizeConfig={{ sizes, maxHeight }}. So the native side receives them as a bundle but the user can just use seperate size and maxHeight props on the component. Thanks @mikehardy for the idea.

Yes, I was going to do this with `width` in there as well.

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsCommon.java
```diff
@@ -51,8 +52,14 @@ static AdSize getAdSizeForAdaptiveBanner(String preDefinedAdSize, ViewGroup reac
...
       display.getMetrics(outMetrics);
       int adWidth = (int) (outMetrics.widthPixels / outMetrics.density);
+      float maxAdHeight = ((ReactNativeAdView)reactViewGroup).getMaxAdHeight();
 
       if ("INLINE_ADAPTIVE_BANNER".equals(preDefinedAdSize)) {
+    if (maxAdHeight > 0) {
+        return AdSize.getInlineAdaptiveBannerAdSize(adWidth, Math.round(Math.max(maxAdHeight, 32)));
+    }
+    return AdSize.getCurrentOrientationInlineAdaptiveBannerAdSize(
+        reactViewGroup.getContext(), adWidth);
```
dylancom, line 62:
```suggestion
```

dylancom (changes requested):

markwilcox:
Update: Still working on this (slowly). Got the iOS new architecture version working with sizes and maxHeight passed as an object (not width yet). Haven't tested old architecture yet, haven't updated Android so won't push the changes as it'll break the branch.

I'm out of office until January, so will probably not get back to finish it until then.

mikehardy:
Happy holidays - hope you enjoy the break @markwilcox !

dylancom:
@markwilcox do you plan to still pick this up?

dylancom:
No problem, just checking in if you still plan to get back to this 😃.

SumitR9910:
@markwilcox Hey just wanted to know if you are still working on this PR.😄

markwilcox:
> @markwilcox Hey just wanted to know if you are still working on this PR.😄

Yes, sorry it keeps getting bumped down the priority list, but we’re going to expand the use of these size restricted banners in our app soon, so I’ll try to get to it in the next 2 weeks.

SumitR9910:
> > @markwilcox Hey just wanted to know if you are still working on this PR.😄
> 
> Yes, sorry it keeps getting bumped down the priority list, but we’re going to expand the use of these size restricted banners in our app soon, so I’ll try to get to it in the next 2 weeks.

Oh great, thanks for the update!

markwilcox:
Updated this PR @SumitR9910 @dylancom

Sorry it has taken a long while. Work has been spread out across a lot of short sessions, so I won't be at all offended about any consistency nitpicks!

Seems to be working for me both platforms and old + new architecture in the test app. I'm going to try running the fork in our production app soon (currently old arch - planning a new arch migration very soon).

ios/RNGoogleMobileAds/RNGoogleMobileAdsBannerView.mm
```diff
@@ -57,18 +57,28 @@ - (void)updateProps:(Props::Shared const &)props oldProps:(Props::Shared const &
...
+      oldViewProps.sizeConfig.maxHeight != newViewProps.sizeConfig.maxHeight ||
+      oldViewProps.sizeConfig.width != newViewProps.sizeConfig.width) {
+    NSMutableArray *adSizes = [NSMutableArray arrayWithCapacity:newViewProps.sizeConfig.sizes.size()];
+    CGFloat maxAdHeight = -1;
+    CGFloat maxAdWidth = 10000; // bigger than any screen width, which will cap it
+    if (newViewProps.sizeConfig.maxHeight > 0) {
+      maxAdHeight = newViewProps.sizeConfig.maxHeight;
+    }
+    if (newViewProps.sizeConfig.width > 0) {
+      maxAdWidth = newViewProps.sizeConfig.width;
```
dylancom, line 70: Why do we switch here to also use the "max" convention for the width? The native methods just use the concept of a width and only **maxHeight** `GADInlineAdaptiveBannerAdSizeWithWidthAndMaxHeight(CGFloat width, CGFloat maxHeight)`.

dylancom:
@markwilcox No problem and thanks again for the contribution! Hope @mikehardy can also have a look.

ios/RNGoogleMobileAds/RNGoogleMobileAdsBannerView.mm
markwilcox: Oh, I just called it max here because the utility method will then constrain it to the screen width. So technically you can set whatever you want, but it'll never be bigger than the screen. I'd be completely fine with just calling it width or adWidth if you prefer?

I think I just went with adWidth on Android... I discovered naming view group properties there as height or width overrides a native view size with some undesirable results - e.g. if the ad fails to load or if the ad returned is smaller than the max height requested it still takes up all the space.

dylancom: I think it will be simpler if we stick with just "width" as defined by the admob API. We don't have to introduce a new maxAdWidth cap of 10000; And in `(GADAdSize)stringToAdSize:(NSString *)value withMaxHeight:(CGFloat)maxHeight andWidth:(CGFloat)width` use width if it was set, otherwise fallback to the screenwidth which we default to now.

ios/RNGoogleMobileAds/RNGoogleMobileAdsBannerComponent.m
```diff
@@ -64,17 +64,28 @@ - (void)setUnitId:(NSString *)unitId {
   _propsChanged = true;
 }
 
-- (void)setSizes:(NSArray *)sizes {
+- (void)setSizeConfig:(NSDictionary *)sizeConfig {
+  NSArray *sizes = sizeConfig[@"sizes"];
   __block NSMutableArray *adSizes = [[NSMutableArray alloc] initWithCapacity:sizes.count];
+  id heightWrapper = sizeConfig[@"maxHeight"];
```
dylancom, line 70: Can be reduced to two lines:
```
CGFloat maxHeight = sizeConfig[@"maxHeight"] ? [sizeConfig[@"maxHeight"] doubleValue] : -1;
CGFloat maxWidth = sizeConfig[@"width"] ? [sizeConfig[@"width"] doubleValue] : 10000;
```

ios/RNGoogleMobileAds/RNGoogleMobileAdsBannerView.mm
markwilcox: Updated this to get rid of the max width cap and treat width more like the max height. Should be simpler now and more consistent with Android.

ios/RNGoogleMobileAds/RNGoogleMobileAdsBannerComponent.m
markwilcox: Updated to use ternary operators along with the above change to not use max width or the arbitrary cap.

src/types/BannerAdProps.ts
```diff
@@ -48,6 +48,18 @@ export interface BannerAdProps {
...
+  /**
+   * Limit inline adaptive banner height.
+   * By default, inline adaptive banners instantiated without a maxHeight value have a maxHeight equal to the device height.
+   */
+  maxHeight?: number;
+
+  /**
+   * Fix inline adaptive banner width.
+   * By default, inline adaptive banners instantiated without a width value have a width equal to the device width.
+   */
```
dylancom, line 60: "Fix" could make it slightly confusing.
This by the way also now sets the width for adaptive banners right?

Suggestion:

```
/**
 * Sets the width for adaptive banners (inline and anchored).
 * If not specified, the width defaults to the full device width.
 */
```

markwilcox: Yes, updated.

Technically the banner might be smaller than the width (it's also effectively a max width) but that's quite unusual - I don't know for sure if that happens on AdMob banners themselves or only some of our mediation sources.

I think this is a clear enough improvement.

dylancom:
Great work @markwilcox, will wait for @mikehardy for assistance on the Android part.

android/src/main/java/io/invertase/googlemobileads/common/ReactNativeAdView.java
```diff
@@ -2,6 +2,7 @@
 
 import android.content.Context;
 import android.widget.FrameLayout;
+import com.facebook.react.bridge.ReadableArray;
```
mikehardy, line 5: this import seems to have been added but there are not other diff `+` lines that reference it ?

android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsCommon.java
```diff
@@ -30,13 +30,15 @@
...
 import com.google.android.gms.ads.admanager.AdManagerAdRequest;
+import io.invertase.googlemobileads.common.ReactNativeAdView;
 import io.invertase.googlemobileads.common.ReactNativeEventEmitter;
 import java.util.ArrayList;
 import java.util.Map;
 import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.annotation.Nullable;
+import android.util.Log;
```
mikehardy, line 41: this import was added but no lines reference it, likely vestige of implementation debugging

dylancom:
@markwilcox Removing the imports and fixing the lint/prettier issues will be enough to merge :)

markwilcox:
@dylancom Imports removed, lint/prettier issues fixed and PR updated to latest main.

markwilcox:
Sorry - Android lint had failed and I'd missed there isn't a lint:android:fix option.

You could just add:
`"lint:android:fix": "google-java-format --replace --glob=\"android/**/*.java\"",`

markwilcox:
Lint should all pass now if you wouldn't mind approving the workflows again @dylancom?

dylancom:
Thanks for contributing @markwilcox, great work!

PR #664, opened 2024-12-01
fix(android): use currentActivity in initialize if !null, aids use of mediation adapters

TaylorDale:
### Description
Required for Unity Ads to function in mediation, as far as I can see shouldn't effect anything else

### Release Summary
Use currentActivity to initialise ads if available

### Checklist

  - [ x] Yes
  - [ x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ x] No
🔥

### DIFF
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsModule.kt
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsModule.kt
@@ -95,6 +95,11 @@ class ReactNativeGoogleMobileAdsModule(
   @ReactMethod
   fun initialize(promise: Promise) {
     MobileAds.initialize(
+      // in react-native, the Activity instance *may* go away, becoming null after initialize
+      // it is not clear if that can happen here without an initialize necessarily following the Activity lifecycle
+      // it is not clear if that will cause problems even if it happens, but users that have widely deployed this
+      // with the use of currentActivity have not seen problems
+      // reference if it needs attention: https://github.com/invertase/react-native-google-mobile-ads/pull/664
       currentActivity ?: reactApplicationContext,
       OnInitializationCompleteListener { initializationStatus ->
         val result = Arguments.createArray()
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsModule.kt
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsModule.kt
@@ -95,7 +95,7 @@ class ReactNativeGoogleMobileAdsModule(
   @ReactMethod
   fun initialize(promise: Promise) {
     MobileAds.initialize(
-      reactApplicationContext,
+      currentActivity ?: reactApplicationContext,
       OnInitializationCompleteListener { initializationStatus ->
         val result = Arguments.createArray()
         for ((key, value) in initializationStatus.adapterStatusMap) {


android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsModule.kt
```diff
@@ -95,7 +95,7 @@ class ReactNativeGoogleMobileAdsModule(
   @ReactMethod
   fun initialize(promise: Promise) {
     MobileAds.initialize(
-      reactApplicationContext,
+      currentActivity ?: reactApplicationContext,
```
mikehardy, line 98: that *seems* innocuous, but I fear there is a lifecycle mismatch between react-native's idea of "the current activity" and the google mobile ads instance initializing here.

Is it possible that the mobile ads instance will maintain hold on this currentActivity and attempt to use it, despite it perhaps being disposed at some point by react-native - making it invalid?

This is the sort thing that will work 99.9% of the time but will fail often enough to really be concerning and require a fix ("activity is null" is a classic react-native problem in lots of modules, I've had to repair it before

Prior art: https://github.com/invertase/react-native-google-mobile-ads/pull/269/files - but funny enough I just checked to see if I could ping the other for results on that one and....it's you! 😆  - so how is it working in practice?

Perhaps...if the initialization were extracted so you could *reinitialize*, and the currentActivity were stored for future-safety-checking, *then* on future methods that could use the initialized MobileAds instance you could see if currentActivity had gone null or was different from the first init and you could re-initialize it with the new-currentActivity?

TaylorDale: Yeah you caught me, it was me last time too haha. In practice I have this deployed to millions of users and I have never really seen any issue I could relate back to it. But completely understand your concerns - I don't really know enough about the Android ecosystem to develop something that is fireproof to what you are talking about here.

This fix is specific and only applies to the unity sdk, so if not possible to merge into main without more changes maybe we just close it out and leave it as an example to anyone else that runs into the issue, if they are willing to try the fix at their own risk?

mikehardy: This:

>  In practice I have this deployed to millions of users over a long period of time and I have never really seen any issue I could relate back to it.

Is a pretty powerful argument that it is sufficient, perhaps in all the cases where the activity lifecycle is also triggered, an initialize is necessarily called after ? 🤔 

But mainly, your observed evidence there is just a good data point

DoctorJohn:
This patch has great timing. I noticed yesterday that upgrading to the latest version results in the Unity Ads adapter not being initialized at all. I even get a fancy error:

```json
{
  "description": {
    "Code": 302,
    "Message": "Unity Ads initialization failed: [INVALID_ARGUMENT] Unity Ads SDK failed to initialize due to invalid context",
    "Domain": "com.unity3d.ads",
    "Cause": "null"
  },
  "name": "com.google.ads.mediation.unity.UnityMediationAdapter",
  "state": 0
}
```

This started occurring after some Android files were recently ported from Java to Kotlin. While I was in the middle of my investigation of the issue, this PR popped up and fixed it. So thanks for that!

mikehardy:
Happy to hear you'll be a test case too @DoctorJohn since you've been active here, and great timing that this came through then. I hope I'm just being too anxious about the "possibly null activity" case but keep your eyes out for any related crashes and let us know. Hopefully we don't see anything but you never know

PR #665, opened 2024-12-04
feat(ios): precision constants on new arch

dylancom:
### Description

Port the precision constants to the new arch.


mikehardy (approved):
clean!

PR #667, opened 2024-12-04
feat(ios): UMP Consent on new arch

dylancom:
### Description

UMP Consent on new arch

_Note that importing types does (still) not work with codegen, so I moved the types to the spec directly._

Any folks that can test this? @TaylorDale @wjaykim @DoctorJohn @birdofpreyru


src/AdsConsent.ts
```diff
@@ -75,53 +71,53 @@ export const AdsConsent: AdsConsentInterface = {
     return native.requestInfoUpdate(options);
   },
 
-  showForm(): Promise<AdsConsentInfo> {
+  showForm() {
```
mikehardy, line 74: interesting, so these are no longer async?
That's a breaking change for people doing old .then/.catch/etc style, but won't matter for people doing async/await (other than lint stuff)

src/specs/modules/NativeConsentModule.ts
```diff
@@ -433,3 +309,215 @@ export interface AdsConsentUserChoices {
...
+   *     // Show a Google-rendered form
+   *     const formResult = await AdsConsent.showForm();
+   *
+   *     console.log('User consent obtained: ', formResult.status === AdsConsentStatus.OBTAINED);
+   *   }
+   * }
+   *
+   * ```
+   */
+  showForm(): Promise<AdsConsentInfo>;
```
mikehardy, line 357: here they are still as async though 🤔 ?

src/AdsConsent.ts
mikehardy: I might just be misreading the type, in that now maybe typescript is inferring things and just infers the type correctly, such that it is async and is typed as described later on where I made my second comment...

mikehardy (approved):
Curious about the async-ness on the type signatures but can't see anything truly wrong
Testing always good though if anyone else has time

src/AdsConsent.ts
dylancom: The type can now be inferred indeed and that's why I removed them.

dylancom:
> Curious about the async-ness on the type signatures but can't see anything truly wrong
> 
> Testing always good though if anyone else has time

Still async. But there is the possibility now to make some methods sync for new arch. Not sure what path we should take there, introduce new methods / breaking change?

src/AdsConsent.ts
dylancom: Types are now defined in the spec (NativeConsentModule.ts) and automatically inferred by the native layer. This is thanks to the use of the React Native Codegen tool, which bridges the gap between TypeScript types and the native implementation.

mikehardy: awesome, resolved on both then - thanks!

PR #669, opened 2024-12-09
chore(ios): remove unused code from old migration

dylancom:
### Description

Remove some unused code that still came from moving of the react-native-firebase lib.


mikehardy (approved):
all the config infrastructure :-), I suppose google mobile ads doesn't need it...or it wouldn't compile

Issue #287, opened 2022-12-08
Parent View with marginHorizontal doesn't affect BannerAd


caiqueex:
### What happened?

Hi all,

Parent View with marginHorizontal doesnt affect BannerAd

`          <View
            style={{
              margin: 16,
              alignItems: 'center',
              justifyContent: 'center',
            }}
          >
            <BannerAd
              unitId={TestIds.BANNER}
              size={BannerAdSize.ANCHORED_ADAPTIVE_BANNER}
            />
          </View>`

I already tested with margin, padding and width and nothing works... And I tested changing prop size in BannerAd too.

does anyone have a suggestion? I want margin because all my screen has 16px of padding/margin. 

![example](https://user-images.githubusercontent.com/8368079/206484885-17c0e3bb-5230-4d2a-9295-b7fbebba8ec0.jpeg)

### Platforms

Android and iOS

### React Native Info

```shell
doesn't apply
```

### Are your using Typescript?

- [X] My project is using Typescript

### package.json

```json
doesn't apply
```

### app.json

```json
doesn't apply
```

### ios/Podfile

```ruby
doesn't apply
```

### android/build.gradle

```groovy
doesn't apply
```

### android/app/build.gradle

```groovy
doesn't apply
```

### android/settings.gradle

```groovy
doesn't apply
```

### AndroidManifest.xml

```xml
doesn't apply
```

mikehardy:
https://github.com/invertase/react-native-google-mobile-ads/blob/443405296804c0e29d17bda3abf162599f6e52ab/src/BannerAdSize.ts#L50-L64

Sounds like you might want FLUID type?

https://github.com/invertase/react-native-google-mobile-ads/blob/443405296804c0e29d17bda3abf162599f6e52ab/src/ads/BaseAd.tsx#L116-L124

Or you could experiment with measuring the container you want to put it in then sending in the size? https://github.com/invertase/react-native-google-mobile-ads/blob/443405296804c0e29d17bda3abf162599f6e52ab/src/types/BannerAdProps.ts#L48

dylancom:
We currently only support device width adaptive banners. But ideally the package should also support setting a custom width.

caiqueex:
> https://github.com/invertase/react-native-google-mobile-ads/blob/443405296804c0e29d17bda3abf162599f6e52ab/src/BannerAdSize.ts#L50-L64
> 
> Sounds like you might want FLUID type?
> 
[...]

PR #674, opened 2024-12-23
feat(banner): allow adaptive banners to use their container's width

youedd:
### Description

In a project I’m working on, I need to apply horizontal padding to adaptive banners for better integration with my app's UI. 

Currently, it's not possible to give adaptive banners a width different from the screen width, as mentioned in the stale issue #287.
If the parent view has a different width, padding, or margin, the banner may not display correctly, and parts of it could be cut off.

This PR adds possibility for adaptive banners to take their container's width, as suggested by @dylancom in https://github.com/invertase/react-native-google-mobile-ads/pull/296#issuecomment-1416857910)

I believe this should be the default behaviour for adaptive banner. but to avoid introducing a breaking change. I added a new prop `adaptiveMode` to enable this functionality:

**- MAIN_SCREEN (default):** Keeps the existing behavior where the banner takes the screen width.
**- CONTAINER:** Allows the adaptive banners to use their container's width, with the ad reloaded when the width changes.
### Related issues

#287
### Release Summary

Allow adaptive banners to use their container's width.
### Checklist

  - [X] Yes
  - [X] `Android`
  - [X] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

### Test Plan

https://github.com/user-attachments/assets/bc2b54a7-d315-4ebc-afd4-6307beed0066

🔥

### DIFF
--- a/RNGoogleMobileAdsExample/App.tsx
+++ b/RNGoogleMobileAdsExample/App.tsx
@@ -185,9 +185,8 @@ class InterstitialTest implements Test {
 class BannerTest implements Test {
   bannerAdSize: BannerAdSize | string;
 
-  constructor(bannerAdSize) {
+  constructor(bannerAdSize: BannerAdSize | string) {
     this.bannerAdSize = bannerAdSize;
-    this.bannerRef = React.createRef();
   }
 
   getPath(): string {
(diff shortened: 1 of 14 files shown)


youedd:
The style changes are removed.
Thanks for the feedback @dylancom !

dylancom:
Thanks again for your contribution! After discussion with the other maintainers, we’ve decided to proceed with #663, as it is already in development and reflects changes we previously agreed upon. We appreciate your effort and encourage you to keep contributing.

mikehardy:
Echoing Dylan - @youedd we really appreciate the contribution, the idea is that #663 *should* also allow you to solve your use case, but goes farther and allows specification of height as well. So if it solves the width use case you need, but also height then it seems better to go with the more comprehensive solution. If you head over to 663 and it seems like it won't let you set width then our assumption is wrong and we should perhaps tune 663 so it can solve your use case

youedd:
My apologies — I saw #663 and assumed it only addressed banner max height before creating this PR.

I'll close this PR, as the discussion on handling width has already started in 663.

I'll just patch the package locally with my changes while waiting for 663 to be merged 🙌

Thank you both for your time, and happy holidays, everyone!

dylancom:
@youedd Do you plan to still pick this up?

PR #814, opened 2025-10-22
build: use yarn@4, update deps, use react-native-test-app for example (w/rn82)

mikehardy:
Gets the example app up to date, and all our internal deps + tooling at the same time

Works for me locally?

### DIFF
--- a/ios/RNGoogleMobileAds/RNGoogleMobileAdsMediaView.h
+++ b/ios/RNGoogleMobileAds/RNGoogleMobileAdsMediaView.h
@@ -30,6 +30,7 @@ NS_ASSUME_NONNULL_BEGIN
 #ifdef RCT_NEW_ARCH_ENABLED
 @interface RNGoogleMobileAdsMediaView : RCTViewComponentView
 
+// this setters/getters here are provided by RCTViewComponentView super, we just narrow the type.
 @property(nonatomic, strong, nullable) GADMediaView *contentView;
 #else
 @interface RNGoogleMobileAdsMediaView : GADMediaView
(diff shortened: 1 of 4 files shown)


src/ads/native-ad/NativeAdContext.ts
```diff
@@ -21,6 +21,6 @@ import type GoogleMobileAdsNativeView from '../../specs/components/GoogleMobileA
 
 type NativeAdContextType = {
   nativeAd: NativeAd;
-  viewRef: RefObject<React.ElementRef<typeof GoogleMobileAdsNativeView>>;
+  viewRef: RefObject<React.ComponentRef<typeof GoogleMobileAdsNativeView> | null>;
```
mikehardy: this line is problematic - the type change is potentially breaking, need to either handle differently or verify not breaking

not sure if types are included in react now but they are in react-native and I don't think I'm using the built-in types yet, I think I'm still using DefinitelyTyped ones

RNGoogleMobileAdsExample/App.tsx
```diff
@@ -199,7 +233,7 @@ class InterstitialTest implements AutoExecutableTest {
 }
 
 class BannerTest implements AutoExecutableTest {
-  bannerRef: RefObject<BannerAd>;
+  bannerRef: RefObject<BannerAd | null>;
```
mikehardy: this appears to be the possible impact of that null change prior as I forward-ported to modern types

src/ads/native-ad/NativeAdContext.ts
mikehardy: @dylancom I can't see a way around this one - our internal logic actually uses it's possible null-ness, I think we just need to issue it as a lightly breaking change.

https://github.com/invertase/react-native-google-mobile-ads/blob/8a5b6ef2abf8b407010ab4cef130caef26b89563/src/ads/native-ad/NativeAsset.tsx#L47-L50

I'm going to re-push with the commit altered so it will trigger the correct breaking change release and notes - I can also *undo* that if you disagree, but you'll get to see what I'll put in there for a final approval pre-merge + release

what do you think?

dylancom:
Fantastic job! Lot's of improvements in there and with a modern example app we can easier adapt the final pieces to the new arch.

PR #854, opened 2026-05-20
fix(native-ads): return image assets and stabilize iOS native ad views

longtimeno-c:
### Description

This PR fixes two native ad issues:

- Returns native ad image assets from Android and iOS via the existing `NativeAd.images` field.
- Improves iOS native ad view stability by resolving the native module through a shared module fallback when the bridge lookup is unavailable.

The TypeScript API already exposes `images: Array<NativeAdImage> | null`, but the native loaders were not populating it. This change fills that native response data on both platforms.

On iOS, `RNGoogleMobileAdsNativeView` and `RNGoogleMobileAdsMediaView` can be initialized in paths where `_bridge moduleForClass:` does not return the native module. Falling back to the shared module instance avoids nil native module access when binding native ads by `responseId`.

### Related issues

N/A

### Release Summary

Native ads now return image assets on Android and iOS, and iOS native ad views are more stable when resolving loaded ads.

### Checklist

  - [ ] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- Applied the change in a local app using native ads.
- Verified iOS native ads no longer crash when rendering native ad/media views.
- Verified native ad image assets are available through `nativeAd.images`.
- Ran `git diff --check`.

Note: TypeScript types already included `NativeAd.images`, so no type shape change was required.

### DIFF
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsNativeModule.kt
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsNativeModule.kt
@@ -67,6 +67,16 @@ class ReactNativeGoogleMobileAdsNativeModule(
       } ?: run {
         data.putNull("icon")
       }
+      val imagesArray = Arguments.createArray()
+      nativeAd.images.forEach { image ->
+        image.uri?.let { uri ->
+          val imageMap = Arguments.createMap()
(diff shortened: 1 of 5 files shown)


mikehardy:
Superceded by #860

PR #874, opened 2026-08-15
fix(android): exclude ad views from instance state saving

philipheinser:
### Description

Banner and native ad views can crash on Android when a fragment (for example `react-native-screens`) restores view hierarchy state.

Mediation adapters such as Facebook Audience Network persist view state under small view ids. Those ids collide with React Native view tags, which produces:

```
Wrong state class, expecting View State but received
com.facebook.ads.internal.util.parcelable.WrappedParcelable
```

This disables instance-state save/restore on the ad view wrappers so parent fragments no longer restore adapter-owned state.

### Related issues

None.

### Release Summary

fix(android): exclude ad views from instance state saving to prevent mediation adapter crashes

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Reproduced by navigating away from a screen that shows a GAM banner or native ad mediated by Facebook Audience Network, then returning so `react-native-screens` restores the fragment view state. Without this change the activity crashes; with `setSaveFromParentEnabled(false)` / `isSaveFromParentEnabled = false` the restore succeeds.

:fire:

### DIFF
--- a/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsNativeAdView.kt
+++ b/android/src/main/java/io/invertase/googlemobileads/ReactNativeGoogleMobileAdsNativeAdView.kt
@@ -43,6 +43,13 @@ class ReactNativeGoogleMobileAdsNativeAdView(
   private var reloadJob: Job? = null
 
   init {
+    // Exclude the ad view hierarchy from instance state saving/restoring. Mediation
+    // adapters (e.g. Facebook Audience Network) save view state under small view ids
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
nice - thanks for this fix @philipheinser - it looks spot on

mikehardy:
ios failure is an unrelated e2e flake, I'll shore that up in an e2e pass shortly. this is good to go

Issue #864, opened 2026-07-09
Google Mobile Ads Next Gen placeholder


mikehardy:
This is just a placeholder to let you know that the maintainers here are aware of the new preferred Android "Next Gen" Google Mobile Ads SDK:

https://ads-developers.googleblog.com/2026/07/announcing-google-mobile-ads-next-gen.html

We're thinking through internally how we can best adapt to this and will have more information in the future

In the meantime, feel free to hit that "subscribe" button in the github UI if you want to get notifications about it from updates on this issue

Cheers

Issue #850, opened 2026-04-26
[📚] Next-Gen SDK


exzos28:
### Documentation Feedback

Hi! I noticed that the Next-Gen SDK is no longer in beta. Are there any plans to migrate?

exzos28:
:(

philippeauriach:
:(

dylancom:
Why don't help this community package instead of waiting for a volunteer to contribute? Easier than ever to submit a PR in the age of AI.

exzos28:
@dylancom Well, I don't mind doing that. I just need to know if a migration is planned.
Why make changes if they aren't going to be merged? Moreover, the answer that this was not planned had already been given a year earlier.

dylancom:
There is no planning, just volunteers submitting PR's of stuff they need.

exzos28:
@dylancom 

I just have two options:
- Explore the current project and try to design a way to use the next-gen SDK in it - with backward compatibility, extensibility, etc. Or maybe a permanent move to next-gen? Lot's of options here.
- Write a separate internal solution using the next-gen SDK right away. Which is much easier.

There's no point in doing PR and putting effort into it, considering that it might not even be merged (! because there was already a response earlier stating that this isn't planned).
And create yet another fork just for the sake of it - come on....

dylancom:
We'd happily merge a PR moving this package forward. Someone back then was requesting the next gen sdk when it was still in beta. I believe the react native user facing API can stay pretty much the same while the android code is updated?

So I'm not sure what background compatibility would be needed.

radekzz:
I think Google said migration is needed, so it should be planned now? 🤔

Issue #836, opened 2026-02-19
[📚] Update to GMA Next-Gen SDK for Android


sbonetpi360:
### Documentation Feedback

Hi,
Our contacts at Google indicate that we should start using the GMA Next-Gen SDK for Android applications now due to its numerous benefits.

https://developers.google.com/ad-manager/mobile-ads-sdk/android/next-gen/quick-start

Although it's in beta, they recommend its use in production environments.
Have you considered making this update?

Thanks,
Sergio.

dylancom:
At the moment we’re not planning to migrate the library to the GMA Next-Gen SDK while it’s still in beta. If you’d like to experiment with it, feel free to fork the repository and try integrating it there.

exzos28:
@dylancom Hi! I noticed that the Next-Gen SDK is no longer in beta. Are there any plans to migrate?

PR #882, opened 2026-08-20
feat(android): add opt-in support for GMA Next-Gen SDK

exzos28:
Adds an opt-in Android build path against Google's newer GMA Next-Gen SDK (ads-mobile-sdk), selected via a `google_mobile_ads_sdk` app.json flag (or `googleMobileAdsSdk` Expo config-plugin parameter). Defaults to the existing legacy SDK (play-services-ads), so existing apps are unaffected.

- Dual native implementations per ad format under `android/src/legacy` and `android/src/nextgen`, selected at build time via a Gradle.
- GMA Next-Gen SDK has no Google Ad Manager equivalent, so Ad Manager ad unit IDs are explicitly rejected under next-gen rather than silently degrading.
- Excludes `play-services-ads` from the classpath when next-gen is selected, since both SDKs ship classes in the same Java package and cannot coexist.
- Exposes `mobileAds().googleMobileAdsSdk` so JS can detect which SDK variant is active (e.g. to hide Ad Manager specific UI).
- Adds Expo config-plugin support, CI build matrix coverage, and docs (docs/next-gen-sdk.mdx).

### Description

Google is pushing a rewritten Mobile Ads SDK for Android as the eventual successor to `play-services-ads`. It's a from-scratch package with no Ad Manager support and its own set of API quirks, so this isn't a drop-in swap — it's wired up as an alternative build target you opt into, with the legacy SDK staying the default.

The two SDKs can't be linked into the same app (same Java package, different classes), which is why this is a build-time choice rather than a runtime one.

### Related issues

https://github.com/invertase/react-native-google-mobile-ads/issues/864
https://github.com/invertase/react-native-google-mobile-ads/issues/850
https://github.com/invertase/react-native-google-mobile-ads/issues/836

### Release Summary

Adds opt-in Android support for Google's GMA Next-Gen SDK via a `google_mobile_ads_sdk` app.json flag.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
  - [x] `jest` tests added or updated in `__tests__`
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan
Built and installed both variants (`legacy` default and `-PgoogleMobileAdsSdk=next-gen`) on a real emulator. Confirmed via logcat that next-gen actually initializes and talks to Google's ad servers, and manually loaded real test ads for interstitial, rewarded, banner and native under both. Ad Manager unit IDs get rejected with a clear error under next-gen; regular AdMob unit IDs for the same formats work normally under both.

`yarn tests:jest` and `yarn tsc:compile` pass. iOS wasn't built locally (no Xcode in this environment) — the only iOS change is a small `constantsToExport` addition, worth having CI confirm.


exzos28:
jet's "Run All" crashes with "Maximum update depth exceeded" on React 19/RN 0.86. Also can't find where jet/mocha actually gets invoked anywhere - no mocha, no config, CI just does react-native run-android. Might be worth moving to Detox at some point?

mikehardy:
Hey @exzos28 👋 ! This is amazing, thank you - also, an amazing coincidence - over @invertase we'd noticed this push for Android next gen now and we've got some time planned starting a couple days ago to not only adopt it here but to give the whole repository a refresh (with Android Next Gen being just one piece)

We've been busy doing analysis on the same but we hadn't published anything yet. Let me see how you've done the work for the NextGen adoption and see what I can get in for you now quickly or what we can harvest - but either way, testing will get a refresh (it's not working well, as you noticed) - and we will have Next Gen support very soon either through this PR, or some other way, really soon

mikehardy:
Hey @exzos28 👋

Thanks for putting this together, and for the genuinely thorough test plan. The overall shape here is what I'd want: build-time selection rather than runtime, legacy staying the default so nobody's upgrade silently changes SDKs, dual source sets, and the classpath exclusion because the two SDKs collide on the same Java package. That last one is a constraint we'd otherwise have found out the hard way, so thank you for working it through.

One correction I think we need before this can land, and I want to be crystal clear about it because it's the load-bearing part. The PR rejects Ad Manager ad unit IDs under next-gen, on the basis that next-gen has no Ad Manager equivalent. According to Google's docs, Ad Manager is supported. There's a full Ad Manager next-gen documentation tree, for example the [quick start](https://developers.google.com/ad-manager/mobile-ads-sdk/android/next-gen/quick-start) and [combining native + banner](https://developers.google.com/ad-manager/mobile-ads-sdk/android/next-gen/native-banner).

If I'm reading it right, what's actually gone is the separate `AdManagerAdRequest` / `AdManagerAdView` classes. Ad Manager folds into the same per-format request builders with Ad Manager-only fields on them (`NativeAdRequest.Builder.addCategoryExclusion` is documented as Ad Manager only, for instance). So I think the fix is to drop the rejection and map the GAM-only options onto the next-gen builders, rather than gating on the shape of the unit ID. Please correct me if I'm wrong, I'd much rather be shown a symbol that doesn't exist than ship a gate that blocks working setups.

This matters more than a normal review nit because Ad Manager publishers are a meaningful share of this library's users, and for them a next-gen path that refuses their unit IDs isn't opt-in, it's unusable.

Fair warning on process, since I don't want you sinking more time into rework blind: we have a larger piece of work going on with this library internally right now, and next-gen Android behind the existing JS API is part of it. I'll come back here with a concrete answer on whether we take this as-is with the Ad Manager fix, take the mechanism and build on it, or land it another way. Either way, the app.json / Expo config flag design and the source-set split are the parts I expect to keep, and I'll credit the work.

One more thing worth flagging because it will bite someone eventually: per Google's [migration guide](https://developers.google.com/admob/android/next-gen/migration), next-gen works only with no mediation or with AdMob as the mediation platform. Apps hosting mediation elsewhere can't use next-gen at all. Probably worth a line in `docs/next-gen-sdk.mdx` if this lands.

Thanks again.

Cheers

exzos28:
Hi @mikehardy

Thanks for the careful read, and yeah, you're right. I went back and checked against the actual SDK rather than assuming, and the rejection was wrong - Ad Manager unit IDs go through the exact same builders as AdMob ones, no separate classes needed. Dropped the gate, wired the GAM-only fields through properly, and un-hid the Ad Manager demos in the example app that I'd gated off for no reason.

Good call on the mediation note too - added a line about that to the docs page.

exzos28:
Quick update - tested on a real emulator against the actual GAM test units:

- GAM Anchored Adaptive Banner - loads fine;
- GAM Interstitial - load/show both work.
- GAM Fluid (reusing `TestIds.GAM_NATIVE` per the existing example comment) - consistently _no-fill_ for me. To rule out a regression, I rebuilt the same example against legacy and got the exact same _no-fill_ on the identical unit/request, so it's not next-gen-specific - likely just inventory/regional availability for that test unit's fluid-shaped creative.

exzos28:
Separately: happy to help set up proper e2e testing for this repo if that's useful, the current jet-based setup seems to have some issues.

Ge0ffreyS:
Hello @mikehardy. Any update on this PR ? Is there anything we can do to help (tests, doc improvement, ...) ? Thanks.

mikehardy:
shipped in #897 - per your note on which ones fill and don't, and how that may skew across classic vs nextgen android, current area of work is automated e2e that covers all API contracts along with deep inspection of any no-fills to make sure there's not some latent error there hiding a real problem behind an inexact no-fill. Already found one of those in product code (and shipped a fix) and in test style (and altered tests). Difficult to get this all right but we should have it provably so shortly. Cheers

mikehardy:
side note: thanks a bunch for the offer of serious help (e2e setup) - just completely redid this repo over the last couple weeks and we've got e2e now. The jet stuff was never quite good enough and I've just gone with good old Appium at the moment

PR #888, opened 2026-09-04
feat: ResponseInfo, errors, lifecycle, multi-format, and ad pools

mikehardy:
## Summary
- Stack slice **1d** (`api-implementation`): FEAT-01…FEAT-06 — meta/ResponseInfo, additive errors, lifecycle ownership, multi-format count-1, classic fullscreen pools, emulated display depth-1 pools.
- Base: `rngma-refactor-infra` (#885). Followed by mediation-adapters and docs PRs.
- User docs / tip-wired API reference land in the docs PR (temporary doc drift is expected).

## Test plan
- [ ] Focus review on ownership/expiry, multi-format mediation safety, and pool peek semantics (see overnight review guide)
- [ ] `yarn tests:jest-coverage` on core
- [ ] Appium smoke both platforms if reviewing native paths

PR #887, opened 2026-09-04
feat(api)!: refreshed API with multiformat and ad pools

mikehardy:
## Summary
- Lands the frozen v17 JS API surface (`feat(api)!`) on top of the OKF kernel branch.
- Includes the OKF session-scratch vs internal gate-state split commit.
- Stack position: **1b** — base `rngma-refactor` (#883), followed by infra / implementation / adapters / docs PRs.

## Test plan
- [ ] Review API freeze / types against `docs/rngma-v17-api-reference.md` (stub-era wording may remain until docs PR)
- [ ] Confirm this PR is reviewed before merging stacked dependents

PR #885, opened 2026-08-26
build(test): yarn workspaces, Appium, coverage, native unit harness

mikehardy:
## Summary
- Stack slice **1c** (trimmed): from `api-design` through pre-FEAT infra only (tip `build(release): update CocoaPods lockfile during release`).
- Yarn workspaces + Lerna/Nx prepare, Appium e2e (Jet removed), `react-native-coverage` adopt, `NativeRNGMATesting` probe, Robolectric/XCTest expansion, release Podfile.lock refresh.
- FEAT / adapters / docs were split onto stacked follow-up branches.

## Test plan
- [ ] `yarn` + `yarn prepare`
- [ ] `yarn tests:appium:validate` (+ device Appium if reviewing e2e harness)
- [ ] `yarn tests:android:unit` / `yarn tests:ios:unit`
- [ ] Confirm base is `api-design` and head stops before `feat(meta):`

PR #883, opened 2026-08-22
docs(okf): agent knowledge kernel for the GMA refactor

mikehardy:
## Summary
- Add a GMA-shaped OKF kernel so agents on this refactor tree follow public vs ephemeral vs private docs, Conventional Commits for commits and PR titles, and gitignored work queues.
- Port only the RNFB patterns that apply here: docs.page maintenance, a `GMA-AD-*` ADR shell, a thin CI triage index, e2e pre-flight/diagnosis gates, and an Expo plugin validation path. Do not copy RNFB monorepo, Detox, TypeDoc, or committed work-queue files.

## Test plan
- [ ] Open `okf-bundle/index.md` on GitHub and confirm relative links resolve (policy, testing, CI, ADRs, docs.page).
- [ ] From `AGENTS.md`, hop to validation/coverage evidence packages and confirm `#fragment` targets exist.
- [ ] Confirm `.agents/work-queues/` and `.agents/reports/` are gitignored and not in the PR diff.
- [ ] Confirm the PR title is Conventional Commits and does not contain an internal tracker id.

---

Maintainer note: Fixes internal CPRN-366

PR #891, opened 2026-09-04
feat(api)!: refreshed API with multiformat and ad pools

mikehardy:
## Summary
- Lands the v17 JS API freeze on `main` (the intended outcome of #887, which merged into `rngma-refactor` instead of `main` after #883).
- Includes the small OKF follow-up that was on the freeze branch but not yet on `main`: split session scratch from internal gate state.
- Same freeze content as amended #887 (durable test filenames; no ephemeral ax* acceptance IDs in public test names).

## Stack note
After this merges, retarget/rebase #885 (`rngma-refactor-infra`) onto the new `main` tip, then the rest of the stack (#888 → #889 → #890) follows.

## Test plan
- [ ] `yarn` / `yarn prepare`
- [ ] `yarn tsc:compile`
- [ ] `yarn tests:jest`
- [ ] `yarn lint:js`
- [ ] Spot-check PR file list: `__tests__/expiryPolicy.test.tsx` (and siblings) present; no `ax*.test.*`


mikehardy (commented):
frozen API agreed on after internal review
marked as breaking but is not breaking yet on this commit - will be breaking before release to bump up required toolchain versions

